PR #243, opened 2026-02-23
fix(naughty): box layout may access nil table

shuber2:
Switching from the VT of somewm causes by_position[s] to be reset to nil. Switching back to it causes a race between re-initializing by_position[s] and handling the signal property::geometry. The latter enumerates by_position[s] when it may be nil.

### DIFF
--- a/lua/naughty/layout/box.lua
+++ b/lua/naughty/layout/box.lua
@@ -139,6 +139,9 @@ end
 -- screen is resized or the notification is moved, it causes side effects.
 -- Better listen to geometry changes and reflow.
 capi.screen.connect_signal("property::geometry", function(s)
+    if not by_position[s] then
+        return
+    end
     for pos, notifs in pairs(by_position[s]) do
         if #notifs > 0 then
             update_position(pos, notifs[1].preset)


shuber2:
With current main (commit d1463d52) I still trigger the race condition.

I added some debug log code and get this when I start somewm, switch to different VT, and then back again:

````
❯ cat /tmp/somewm.log 
1771959614: init_screen(s = screen{index=1, name=eDP-1})
1771959617: Screen removed on scr = screen{index=1, name=eDP-1}
1771959618: property::geometry on s = screen{index=1, name=eDP-1}
1771959618: init_screen(s = screen{index=1, name=eDP-1})
1771959618: init_screen(s = screen{index=1, name=eDP-1})
1771959618: init_screen(s = screen{index=1, name=eDP-1})
1771959622: init_screen(s = screen{index=1, name=eDP-1})
1771959622: init_screen(s = screen{index=1, name=eDP-1})
1771959622: init_screen(s = screen{index=1, name=eDP-1})
````

**Comment by JimmyCozza** [comment]
OK, here's how I understand it so far:


```
// Loop 1: geometry update (fires property::geometry)
  wl_list_for_each(m, &mons, link) {
      luaA_screen_update_geometry(screen);  // ← emits property::geometry
  }

  // Loop 2: screen_added (fires _added → init_screen)
  wl_list_for_each(m, &mons, link) {
      screen_added(screen);  // ← emits _added, which triggers init_screen()
  }
```

  For new screens (needs_screen_added=1), property::geometry fires in loop 1 before _added fires in loop 2. Naughty's handler expects init_screen() to have run, but it hasn't — hence by_position[s] is nil.

  AwesomeWM avoids this because screen_refresh() processes adds before modifications — new screens get screen_added() before any screen_modified().
    
Would you be OK if I open a PR that has some underlying C changes that you could look at?  I don't want to drag this out too long though, since I want to get a real fix in place for you.

Issue #294, opened 2026-02-28
Support touchpad gesture wayland protocol


shuber2:
## Description

Implement the [zwp_pointer_gestures_v1](https://wayland.app/protocols/pointer-gestures-unstable-v1#zwp_pointer_gestures_v1) protocol to enable touchpad gestures like pinch-to-zoom.

## Use Case

Use pinch-to-zoom in browsers like firefox or other applications.

## Proposed Solution

I believe that sway uses the wlroots implementation.

## Alternatives Considered

There are none to my knowledge.

## AwesomeWM Compatibility Note

AwesomeWM could do so, but this is something that X11 does for AwesomeWM I believe.

PR #296, opened 2026-02-28
feat(input): Implement wlr_pointer_gestures_v1 (issue #294)

shuber2:
Implements listeners for swipe, pinch and hold events and passes them on to clients. Enables features like pinch-to-zoom for various applications.

## Description
Implements listeners for the events defined in `wlr_cursor` and passes them straight to the client.

## Test Plan
Tested pinch-to-zoom feature in firefox, gwenview and libreoffice.

## Checklist


JimmyCozza:
@shuber2 

This is great stuff!  I want to design the "full" API for this gestures feature so that it feels kind of like it could be something that would have been part of AwesomeWM.

Let me think about this for a bit, but I think this looks great

Issue #330, opened 2026-03-08
Sizes and floating of dialogs


shuber2:
## Description

Gimp is a quite dialog-intense application. When I start gimp in sway pretty much with default config then it launches like in this screenshot:

<img width="3838" height="2152" alt="Image" src="https://github.com/user-attachments/assets/277ce805-3356-419f-ae94-2c8b29720d58" />

However, with somewm it launches like this:

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/8eb3c962-0534-419e-a87b-822037484270" />

Note that the dialog is not floating, but the client takes up the left half of the tiling layout. It would be great if the somewm default config would create this dialog with `floating=true` by default.

But when we set floating to true then it looks like this:

<img width="1842" height="1161" alt="Image" src="https://github.com/user-attachments/assets/840c168f-7870-4320-8a36-3dff43cec6d2" />

The dialog client is too small and it does not contain all of the content.

Interestingly, if I take the screenshot with `somewm-client` then I cannot show this effect, because magically it shows the correct dialog content:

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/936b918f-eef0-42ec-9d1e-830e1514ca0f" />

## Version Info

```
**somewm:** 0.5.0-190-g1f7a0b5 (2026-03-07)
**wlroots:** 0.19
**Lua:** LuaJIT 2.1.1731601260 (compiled: Lua 5.1.4)
**LGI:** 0.9.2
**Build:** D-Bus=yes, XWayland=yes

**System:**
- Distro: Gentoo Linux
- Kernel: 6.18.12-gentoo-x86_64
- Arch: x86_64
- GPU: amdgpu (1002:15BF)
- Session: wayland (nested: yes)
```

PR #334, opened 2026-03-10
Fix handling dialog-type client windows (#330)

shuber2:
## Description

Three fixes:
1. Let floating windows decide on their own size
2. Set client type to WINDOW_TYPE_DIALOG for floating windows
3. Not sure about this: Symmetrically also call property_register_wayland_listeners() for clients if there is a parent, too.

## Test Plan

Not familiar with the test facilities.

## Checklist


JimmyCozza (approved):
looks good

PR #366, opened 2026-03-21
feat(restart): in-process Lua hot-reload for awesome.restart()

### LINKED ISSUE #232: awesome.restart() crashes (SIGSEGV) — interim cold restart approach
## Problem

`awesome.restart()` crashes with SIGSEGV because `argv` is NULL after `globalconf_init()` calls `memset(&globalconf, 0, ...)`. The original AwesomeWM restart path (`execvp`) also cannot work correctly in a Wayland compositor context — the DRM session, Wayland clients, and GPU state do not survive an exec.

## Interim approach (not a proposed upstream change)

This is **not an enhancement proposal** — it is a practical workaround we are using in our fork for the coming months until a proper solution is designed. Documenting it here for visibility and discussion.

We use the dwm-style exit code pattern:

| Exit code | Meaning | Session script action |
|-----------|---------|----------------------|
| 0 | Normal quit (`awesome.quit()`) | Exit |
| 1 | Cold restart (`awesome.restart()`) | Restart compositor in 1s |
| 2 | Rebuild + restart (`awesome.rebuild_restart()`) | `ninja -C build` then restart |

A `somewm-session` wrapper script runs the compositor in a loop, interprets exit codes, and handles restart/rebuild. All Wayland clients are killed on restart (cold restart) — the OS cleans up everything cleanly.

### What it provides
- Working `awesome.restart()` (no longer crashes)
- Unified session launcher (`somewm-session`) for TTY launch
- Dev workflow: `SOMEWM_DIR=... somewm-session` runs from build dir without sudo
- `somewm-client restart` / `somewm-client rebuild` IPC commands

### Implementation
- `cold_restart()` / `rebuild_restart()` in `somewm.c` — set exit code, call `some_compositor_quit()`
- `main()` returns `globalconf.exit_code` instead of `EXIT_SUCCESS`
- `globalconf_init()` saves `argc`/`argv`/`log_level` before `memset` (real bug fix regardless of approach)

Branch: https://github.com/raven2cz/somewm/tree/feat/cold-restart

## Questions for upstream

1. Is a proper in-process hot restart planned? (preserving client connections like Sway's `swaymsg reload`)
2. Should `awesome.restart()` be documented as unsupported in somewm for now?
3. Would a session wrapper script like `somewm-session` be useful upstream?

JimmyCozza:
## Description

In-process Lua hot-reload for `awesome.restart()`. Tears down the Lua VM and rebuilds it from rc.lua while Wayland clients survive. wlroots, the scene graph, and client surfaces are untouched.

`client_t` objects are the hard part: they're Lua userdata with ~20 embedded `wl_listener` fields linked into wlroots signal chains. The reload snapshots them to the heap, detaches listeners, creates fresh userdata in the new state, and reattaches. The `::connected` pattern in luaclass.c handles retroactive screen initialization so screens just need to exist before rc.lua loads.

The old Lua state is intentionally leaked (~1-2 MB) to avoid Lgi closure crashes. A companion LD_PRELOAD library (`lgi_closure_guard.so`) interposes `ffi_prep_closure_loc` to block stale FFI closures from firing after reload. The guard uses versioned symbol interposition, `dladdr()`-based callback detection, and pre-dispatch Lua registry validation. somewm auto-detects and loads the guard via re-exec at startup; `LD_PRELOAD` is cleared before forking children.

Thanks to @raven2cz for tracking down the Lgi FFI closure SEGV root cause and building the initial closure guard implementation.

Closes #232

### Looking for testers

This works well enough to daily-drive. Grab the latest `feat/hot-reload` and try a few reloads (`somewm-client reload` or bind `awesome.restart()` to a key). Please include `/tmp/somewm.log` output with any reports.

### Known issues

- [x] **Systray icons disappear** - ~~Icons disappeared after reload until the SNI watcher re-enumerated.~~ Fixed: systray items are now snapshotted and re-probed after reload, with GDBus singleton cache bypass.
- [x] **Delayed Lgi closure crash** - ~~Old Lgi/libffi closures from the leaked state could fire and SEGV minutes after reload.~~ Mitigated: LD_PRELOAD closure guard blocks stale FFI closures. GLib source sweep removes Lua-owned sources before teardown.
- [x] **Duplicate wibars** - ~~Old drawin scene trees persisted after reload, causing visible duplicates.~~ Fixed: old drawin scene trees and shadow textures are destroyed before leaking the old state.
- [x] **Log noise after reload** - "Reference not found" warnings and "error in exit callback" messages from old Lua state callbacks. Harmless, no crash.
- [ ] **dbus_proxy library** - Using `dbus_proxy` (lgi D-Bus wrapper) causes `(error object is not a string)` on reload. Likely related to GDBus singleton cache invalidation.
- [x] **System freeze** - Reported on some hardware after multiple reloads. Mouse/keyboard/VT switching all lock up. Under investigation (see @shuber2's reports).
- [x] **Lgi closure guard is a workaround** - The LD_PRELOAD guard prevents the SEGV but Lgi's closure lifecycle has no proper teardown API (see lgi-devs/lgi#133, lgi-devs/lgi#9), so the guard is the practical long-term solution for hot-reload.

### DIFF
--- a/luaa.c
+++ b/luaa.c
@@ -205,6 +205,9 @@ static IdleTimeout idle_timeouts[MAX_IDLE_TIMEOUTS];
 static int idle_timeout_count = 0;
 static bool user_is_idle = false;    /* Global idle state */
 
+/* Flag whether idle timers have been inhibited. */
+static bool idle_timers_inhibited = false;
+
 /* D-Bus library functions from dbus.c */
 extern const struct luaL_Reg awesome_dbus_lib[];
 
@@ -1407,6 +1410,35 @@ remove_idle_timeout_at(int idx)
 	idle_timeout_count--;
 }
 
+static void
+reset_idle_timer(IdleTimeout *timeout) {
+	timeout->fired = false;
+	if (timeout->timer) {
+		/* Disable timer if idle timers are inhibited. */
+		int ms_delay = idle_timers_inhibited ? 0 : timeout->seconds * 1000;
+		wl_event_source_timer_update(timeout->timer, ms_delay);
+	}
+}
+
+/* Update all idle timers, i.e., update timer with 0 or timeout depending on
+ * inhibition. */
+static void
+reset_all_idle_timers()
+{
+	for (int i = 0; i < idle_timeout_count; i++) {
+		IdleTimeout *timeout = &idle_timeouts[i];
+		reset_idle_timer(timeout);
+	}
+}
+
+/* Set inhibit state for all idle timers and reset timers. */
+void
+some_idle_timers_set_inhibit(bool inhibit)
+{
+	idle_timers_inhibited = inhibit;
+	reset_all_idle_timers();
+}
+
 /** awesome.set_idle_timeout(name, seconds, callback)
  * Add or update a named idle timeout.
  * Multiple timeouts can be active simultaneously.
@@ -1441,11 +1473,10 @@ luaA_awesome_set_idle_timeout(lua_State *L)
 	timeout->name = strdup(name);
 	timeout->seconds = seconds;
 	timeout->lua_callback_ref = callback_ref;
-	timeout->fired = false;
 
 	/* Create and arm the timer */
 	timeout->timer = wl_event_loop_add_timer(some_get_event_loop(), idle_timeout_callback, timeout);
-	wl_event_source_timer_update(timeout->timer, seconds * 1000);
+	reset_idle_timer(timeout);
 
 	idle_timeout_count++;
 
@@ -1531,12 +1562,8 @@ some_notify_activity(void)
 	}
 
 	/* Reset all idle timers */
-	for (int i = 0; i < idle_timeout_count; i++) {
-		IdleTimeout *timeout = &idle_timeouts[i];
-		timeout->fired = false;
-		if (timeout->timer)
-			wl_event_source_timer_update(timeout->timer, timeout->seconds * 1000);
-	}
+	if (!idle_timers_inhibited)
+		reset_all_idle_timers();
 }
 
 /* ==========================================================================
@@ -1804,7 +1831,7 @@ luaA_awesome_index(lua_State *L)
 	}
 
 	if (A_STREQ(key, "idle_inhibited")) {
-		lua_pushboolean(L, some_is_idle_inhibited());
+		lua_pushboolean(L, some_is_idle_inhibited(NULL));
 		return 1;
 	}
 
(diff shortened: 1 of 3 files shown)


raven2cz:
## Testing hot-reload + fix for Lgi FFI closure SEGV

I tested this branch on our fork (NVIDIA RTX 5070 Ti, Arch Linux, LuaJIT). Hot-reload works great for the first reload — config changes apply instantly, clients survive. However, we hit a consistent SEGV after the second reload:

```
SEGV on unknown address 0x000000000018
    #0 corelgilua51.so  (lgi FFI closure callback)
    #1 libffi.so
    #2 libglib-2.0.so  g_main_loop_run
```

### Root cause

After hot-reload, old Lgi FFI closures remain registered as GLib source callbacks. They hold `lua_State*` pointers to the leaked old Lua VM. When GLib dispatches them (typically 30-60 seconds later via a timer or dbus callback), `lua_status(NULL)` → SEGV.

The DBus close in Phase C prevents dbus-specific callbacks, but Lgi also registers non-dbus GLib sources: timers (`g_timeout_add`), idle handlers (`g_idle_add`), spawn I/O watchers (`g_io_add_watch`). These survive the reload.

### Fix: GLib source sweep by ID baseline

GLib source IDs are monotonically increasing. We record the highest ID after compositor setup (`glib_source_baseline`). Before hot-reload teardown:

1. Cancel compositor-owned sources above baseline (`activation_token_timeout`)
2. Close GDBus connection
3. Probe current ID for exact upper bound
4. Sweep all sources in `[baseline+1, probe_id)` — these are Lua-owned
5. Update baseline for next reload

This eliminates all stale GLib sources before the old Lua state is abandoned.

### Results

- **First reload**: stable, no crash ✅
- **Second reload**: still crashes — leaked FFI closures from first state persist in `ffi_closure` memory (allocated outside Lua heap via `ffi_closure_alloc`)
- `lua_close()` on old state would free them but crashes because client snapshots still reference Lua userdata

### Known limitation (upstream Lgi)

This is a [known unsolved problem](https://github.com/lgi-devs/lgi/issues/133) in Lgi. GLib/GObject [has no deinit API](https://github.com/lgi-devs/lgi/issues/9). [Way-cooler hit the same wall](https://github.com/way-cooler/way-cooler/issues/555) and concluded *"Thou Shalt Not Drop The Lua without getting rid of the whole process"*.

Possible future solutions:
- Isolate Lua/Lgi work onto a separate `GMainContext` that can be destroyed wholesale
- Run Lua in a child process with IPC

### Branch

[`raven2cz/somewm:fix/hot-reload-lgi-crash`](https://github.com/raven2cz/somewm/tree/fix/hot-reload-lgi-crash) — commit https://github.com/raven2cz/somewm/commit/7fe1a73

The source sweep alone makes hot-reload much more stable for daily use (single reload per session is the common case). The second-reload crash is an Lgi architectural limitation that would need upstream changes to fully resolve.

PR #389, opened 2026-03-30
fix: Respect idle inhibitor in idle timeout handler

shuber2:
The some_notify_activity() is called directly in somewm upon user activity. This resets the idle timer as expected. For more elaborate idle inhibition, somewm implements the idle-inhibit-unstable-v1 protocol, which enables tools like wayland-pipewire-idle-inhibit to inhibit the idle mechanism on more complex conditions, e.g., when pipewire plays media.

While checkidleinhibitor() did check for this inhibitor mechanism to call wlr_idle_notifier_v1_set_inhibited(), these checks were not employed when firing the idle::start signal in idle_timeout_callback().

Fix this:

1. Factor the inhibitor check out from checkidleinhibitor() and put it into a generalized some_is_idle_inhibited().

2. Check some_is_idle_inhibited() within idle_timeout_callback().

3. Make some_is_idle_inhibited() robust against idle_inhibit_mgr being NULL.

## Checklist

### DIFF
--- a/luaa.c
+++ b/luaa.c
@@ -1770,7 +1770,7 @@ luaA_awesome_index(lua_State *L)
 	}
 
 	if (A_STREQ(key, "idle_inhibited")) {
-		lua_pushboolean(L, some_is_idle_inhibited());
+		lua_pushboolean(L, some_is_idle_inhibited(NULL));
 		return 1;
 	}
 
--- a/luaa.c
+++ b/luaa.c
@@ -192,6 +192,9 @@ static IdleTimeout idle_timeouts[MAX_IDLE_TIMEOUTS];
 static int idle_timeout_count = 0;
 static bool user_is_idle = false;    /* Global idle state */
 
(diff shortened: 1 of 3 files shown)


JimmyCozza (commented):
The some_is_idle_inhibited() refactor is definitely an improvement.  Let's keep that.

However, for the rest of the fix, what do you think about handling this at the inhibitor lifecycle level instead of polling at the timer fire time?  `createidleinhibitor` and `destroyidleinhibitor` already exist as event hooks and they could pause/resume the idle timers directly.

So something like

- Inhibitor created (and now effectively inhibited): disarm all idle timers with wl_event_source_timer_update(timer, 0)
- Inhibitor destroyed (and no longer inhibited): restart all timers from their full duration

That would prevent the timers from re-firing evern timeout->seconds while inhibited, and then when the inhibitor is removed, the countdown starts immediately rather than being delayed (up until the full interval, possibly).

JimmyCozza (approved):
👽

PR #421, opened 2026-04-07
fix: deliver pointer enter to newly mapped layer surfaces

raven2cz:
## Summary

When a layer-shell surface maps under a stationary cursor, it does not receive `wl_pointer.enter` until the user physically moves the mouse. This breaks hover interactions on panels/overlays that appear under the cursor (e.g. wlr-layer-shell panels triggered by screen edge hover).

## Root cause

`commitlayersurfacenotify()` detects layer surface map transitions (`was_mapped=0 → l->mapped=1`) but does not re-evaluate pointer focus afterward. The cursor remains focused on whatever was under it before the surface appeared.

Sway handles this via `cursor_rebase_all()` in `handle_map()` (`sway/desktop/layer_shell.c`). somewm's own `unmaplayersurfacenotify()` already calls `motionnotify(0, NULL, 0, 0, 0, 0)` for the reverse case (unmap).

## Fix

Add `motionnotify(0, NULL, 0, 0, 0, 0)` after `arrangelayers()` when a layer surface transitions from unmapped to mapped.

```c
if (!was_mapped && l->mapped && !exclusive_focus)
    motionnotify(0, NULL, 0, 0, 0, 0);
```

## Design decisions

- **After `arrangelayers()`**: scene node geometry must be updated before `xytonode()` hit-test
- **`!exclusive_focus` guard**: prevents disrupting keyboard grabs (e.g. session lock)
- **`time=0`**: only runs `xytonode()` + `pointerfocus()` for `wl_pointer` delivery, skips Lua mouse signals
- **`CurPressed` safe**: `motionnotify()` keeps focus on drag target when button is held

Regular clients (XDG/XWayland) are not affected — `mapnotify()` already has a cursor-in-geometry check with direct `pointerfocus()` call.

## Test plan

- [ ] Layer surface maps under stationary cursor → receives `wl_pointer.enter` immediately
- [ ] Layer surface maps while `exclusive_focus` is active → no disruption
- [ ] Layer surface maps while mouse button held → no focus steal mid-drag
- [ ] Layer surface unmaps under cursor → focus returns to surface below (existing behavior, unchanged)
- [ ] Multi-output: layer surface maps on non-focused output → no spurious pointer events

### DIFF
--- a/protocols.c
+++ b/protocols.c
@@ -218,6 +218,16 @@ commitlayersurfacenotify(struct wl_listener *listener, void *data)
 
 	arrangelayers(l->mon);
 
+	/* When a layer surface maps, re-evaluate pointer focus so that
+	 * wl_pointer.enter is delivered if the cursor is already over it.
+	 * Without this, hover doesn't work until the user moves the mouse.
+	 * Similar to sway's cursor_rebase_all() in handle_map(), but uses
+	 * motionnotify(0,...) which re-runs xytonode() + pointerfocus()
+	 * without emitting Lua mouse signals (those are gated on time != 0).
+	 * Skip when exclusive_focus is active to avoid disrupting grabs. */
+	if (!was_mapped && l->mapped && !exclusive_focus)
+		motionnotify(0, NULL, 0, 0, 0, 0);
+
 	/* Emit property change signals for Lua (matches AwesomeWM pattern) */
 	if (l->lua_object && globalconf_L && layer_surface->current.committed) {
 		lua_State *L = globalconf_get_lua_State();


raven2cz:
Rebased onto current `upstream/main` (`d27fa2b`), force-pushed `ceba061`.

**Note:** the upstream refactor moved `commitlayersurfacenotify` from `somewm.c` into `protocols.c` (kolo6 split), which is why the original commit was marked CONFLICTING. The fix itself is a single 10-line addition to the same function, just at the new location.

**Relation to `901e363` (upstream, @JimmyCozza):** complementary, not overlapping. That commit adds `motionnotify(0,...)` after `banning_refresh()` to cover the *tag-switch unban* case. This PR covers the *new layer-surface map* case — a panel appearing under a stationary cursor. Both paths need it; neither subsumes the other.

**Testing:**
- `ninja -C build-test` clean
- Live-tested on NVIDIA RTX 5070 Ti DRM session: launching a Quickshell panel or `swaync` notification under a stationary cursor now triggers hover immediately, without requiring a mouse jiggle. No regressions in session lock (exclusive_focus guard holds).

JimmyCozza:
got these on both release/1.4 and main.  thanks.

PR #427, opened 2026-04-07
fix(stack): place override_redirect XWayland surfaces in overlay layer

### LINKED ISSUE #415: Wine menus appear below the application
See screenshot: Image

If the application is maximized, the menus are effectively hidden (but I can still click on the invisible menu entries). I can also drag the application around and the menu will stay in the same place. Confirmed that this issue doesn't happen on Wayland KDE Plasma.

## Version Info

```
## somewm version info

**somewm:** 0.5.0-269-g4eff245 (2026-04-03)
**wlroots:** 0.19
**Lua:** LuaJIT 2.1.1774896198 (compiled: Lua 5.1.4)
**LGI:** 0.9.2
**Build:** D-Bus=yes, XWayland=yes

**System:**
- Distro: CachyOS
- Kernel: 6.19.11-1-cachyos
- Arch: x86_64
- GPU: amdgpu (1002:13C0)
- Session: wayland (nested: yes)
```

Sidenote, just noticed that `amdgpu` (I'm guessing integrated graphics) is shown as the GPU instead of my Nvidia 4070S. Is that, er, normal?

## Configuration

Config: https://github.com/sig-kill/dots/tree/main/config/somewm/.config/somewm. The only relavent rule I have on Wine windows is to add the titlebar: https://github.com/sig-kill/dots/blob/51da1b84d513e20f63bd7f0164b45f66ed951c2f/config/somewm/.config/somewm/rules.lua#L55-L61

```lua
  ruled.client.append_rule {
    id         = "titlebars",
    rule_any   = {
      class    = { "Wine", "wine" },
    },
    properties = { titlebars_enabled = true }
  }
```

raven2cz:
## Description

Override_redirect X11 surfaces (Wine menus, Steam popups, tooltips) appear **below** their parent window instead of above it.

### Root cause

`stack_refresh()` calls `client_layer_translator()` on **all** clients in `globalconf.stack`, including unmanaged (override_redirect) ones. Since unmanaged clients don't have `floating`, `ontop`, or other stacking attributes set, `client_layer_translator()` returns `WINDOW_LAYER_NORMAL` → `LyrTile`. This undoes the `LyrFloat` placement from `mapnotify()`, moving popups below floating parent windows.

### Fix

Two changes:

1. **`stack.c` — `client_layer_translator()`**: Return `WINDOW_LAYER_ONTOP` for override_redirect clients, mapping them to `LyrOverlay`. This matches X11 semantics where override_redirect windows bypass the window manager and display above all managed windows. Analogous to Sway's separate `root->layers.unmanaged` layer.

2. **`somewm.c` — `mapnotify()`**: Place unmanaged clients directly in `LyrOverlay` (was `LyrFloat`) for consistency with `stack_refresh()`.

### Commit

- [`005242b`](https://github.com/raven2cz/somewm/commit/005242b) — fix + test + override_redirect test helper

Fixes #415

## Test Plan

- [x] **foobar2000 (Wine)**: Right-click context menus and dropdown menus now appear above the application window. Previously appeared completely hidden underneath.
- [x] **Akonadi Console (Qt/XWayland)**: Popup menus display correctly above the application.
- [x] Managed client stacking unaffected (floating, fullscreen, ontop, transients all work)
- [x] Session lock still covers override_redirect surfaces (`LyrBlock` > `LyrOverlay`)
- [x] Integration test: `tests/test-xwayland-override-redirect-stacking.lua` — spawns managed X11 client + override_redirect window, toggles stacking properties (ontop, above, floating, fullscreen), verifies compositor survives
- [x] Test helper: `tests/helpers/x11_override_redirect.py` — creates override_redirect X11 window via libX11 ctypes for test use
- [x] Build passes (`make build-test`)

**Hardware tested:** NVIDIA RTX 5070 Ti (proprietary driver, Arch Linux, DRM session). Not yet tested on RTX 4070 S, but the fix is GPU-independent (pure scene graph layer logic) so behavior should be identical.

## Checklist
- [x] Tests pass (`make build-test`)

### DIFF
--- a/stack.c
+++ b/stack.c
@@ -17,6 +17,9 @@
 #include "somewm_api.h"
 #include <stdbool.h>
 #include <wlr/types/wlr_scene.h>
+#ifdef XWAYLAND
+#include <wlr/xwayland.h>
+#endif
 
 /* Flag to mark stack as needing refresh */
 static bool need_stack_refresh = false;
@@ -226,6 +229,19 @@ stack_refresh(void)
 	foreach(node, globalconf.stack) {
 		if (!(*node) || !(*node)->scene)
 			continue;
+
+		/* Unmanaged (override_redirect) X11 clients have no stacking
+		 * attributes (ontop, floating, fullscreen, ...) and must not be
+		 * reparented out of the layer mapnotify() placed them into
+		 * (LyrOverlay). Running them through client_layer_translator()
+		 * returns WINDOW_LAYER_NORMAL (LyrTile) by default, which
+		 * drops Wine/Steam/Qt popups below their floating parents.
+		 * Inlined check (client.h has cross-file dependencies). */
+#ifdef XWAYLAND
+		if ((*node)->client_type == X11 &&
+		    (*node)->surface.xwayland->override_redirect)
+			continue;
+#endif
 
 		layer = client_layer_translator(*node);
 
--- a/tests/helpers/x11_override_redirect.py
+++ b/tests/helpers/x11_override_redirect.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python3
+"""X11 test helper: create an override_redirect window (simulates popup/menu).
+
+Usage: python3 x11_override_redirect.py <WM_CLASS> [x y width height]
+
+Creates an override_redirect=True window, which bypasses the window manager.
+This is what Wine, Steam, and other X11 apps use for popup menus and tooltips.
+
+The window stays mapped until SIGTERM is received.
+"""
+
+import ctypes
+import ctypes.util
+import signal
+import sys
+
+# --- Load libX11 via ctypes ---
+
+_x11_path = ctypes.util.find_library("X11")
+if not _x11_path:
+    print("ERROR: libX11 not found", file=sys.stderr)
+    sys.exit(1)
+
+x11 = ctypes.cdll.LoadLibrary(_x11_path)
+
+# Type aliases
+Display_p = ctypes.c_void_p
+Window = ctypes.c_ulong
+Atom = ctypes.c_ulong
+
+# XSetWindowAttributes structure (partial — only fields we need)
+class XSetWindowAttributes(ctypes.Structure):
+    _fields_ = [
+        ("background_pixmap", ctypes.c_ulong),
+        ("background_pixel", ctypes.c_ulong),
+        ("border_pixmap", ctypes.c_ulong),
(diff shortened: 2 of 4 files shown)


stack.c
```diff
@@ -85,6 +85,14 @@ client_layer_translator(Client *c)
 	if (!c)
 		return WINDOW_LAYER_NORMAL;
 
+	/* Override-redirect (unmanaged) X11 surfaces are always on top.
+	 * X11 semantics: override_redirect bypasses the WM entirely and
+	 * displays above all managed windows. Maps to LyrOverlay. */
+#ifdef XWAYLAND
+	if (c->client_type == X11 && c->surface.xwayland->override_redirect)
```
JimmyCozza: Should this be a client_is_unmanaged() skip in stack_refresh() instead? Unmanaged clients don't have stacking attributes, so running them through the layer translator at all seems like the wrong boundary.

raven2cz:
Refactored per your feedback, @JimmyCozza — thank you, your approach is cleaner. Force-pushed `ba17507`.

**Changes:**

1. **`stack.c`**: removed the X11/override_redirect special case from `client_layer_translator()`. Added an inline `client_is_unmanaged` check at the top of the `stack_refresh()` foreach loop that `continue`s before the translator is called. Unmanaged clients bypass the stacking model entirely now.
   - Inlined the override_redirect check (wrapped in `#ifdef XWAYLAND`) rather than including `client.h`, since `client.h` pulls in cross-file dependencies (`seat`, `wlr_fractional_scale_v1_notify_scale`) that aren't declared in `stack.c`'s context.

2. **`window.c` (`mapnotify()`)**: kept the `LyrFloat` → `LyrOverlay` change; updated the comment to note that `stack_refresh()` now preserves this placement.

Branch rebased from the stale pre-refactor base onto current `upstream/main` (`d27fa2b`). The commit is a single clean delta on top of HEAD.

**Testing:**
- `make build-test` clean
- `tests/test-xwayland-override-redirect-stacking.lua` passes (1.53s) — exercises the `stack_refresh` path via toggled `ontop`/`above`/`floating`/`fullscreen` on a managed client while an override_redirect window is present
- Live-tested on NVIDIA RTX 5070 Ti DRM session: Wine right-click menus and Qt/XWayland dropdowns appear above their parents, no regressions in managed client stacking or session lock coverage.

JimmyCozza:
Got this merged into release/1.4 and main.  thanks

PR #448, opened 2026-04-12
Feat/2.0 event queue

JimmyCozza:
## Description

> **TL;DR:** Stop interrupting the compositor mid-frame to run Lua signal handlers. Queue them up and run them all at a clean frame boundary instead.

### Problem

The synchronous signal dispatch model we inherited from AwesomeWM (and still ship in 1.4.0) causes three problems:

1. Signal handlers calling into signal handlers calling into signal handlers makes frame timing unpredictable.
2. The C/Lua boundary gets crossed dozens of times per frame when once would do.
3. C and Lua are interwoven together instead of cleanly separated, so "the compositor does work" and "the user config decides policy" happen in the same call stack, interleaved.

### Hypothesis

Batching signal emissions into a queue and draining them at well-defined frame boundaries should measurably improve all three.

### A concrete example

A single `client_resize_do()` can fire seven property signals back-to-back, and a layout pass touching N clients multiplies that. One `awful.layout.arrange()` call can easily cause 30+ C→Lua round-trips before any rendering happens.

**Before:**

```text
C: resize A ─→ Lua (7 handlers) ─→
C: resize B ─→ Lua (7 handlers) ─→
C: resize C ─→ Lua (7 handlers) ─→
C: resize D ─→ Lua (7 handlers) ─→
C: resize E ─→ Lua (7 handlers) ─→
C: apply to scene graph
C: render frame
```

After:

```
C: resize A, B, C, D, E
C: apply to scene graph
           ↓
      [frame boundary]
           ↓
Lua: drain all queued handlers in one pass
C: render frame
```

### What changed

**Signals converted to queued dispatch:**

| Category  | Signals                                                                  |
| --------- | ------------------------------------------------------------------------ |
| Geometry  | `property::geometry`, `property::x/y/width/height`, `property::position`, `property::size` |
| Focus     | `focus`, `unfocus`, `client::focus`, `client::unfocus`, `property::active` |
| Mouse     | `mouse::enter`, `mouse::leave`, `mouse::move` (with coalescing)         |
| Lifecycle | `manage`, `list`, `swapped`                                              |
| Request   | `request::activate`, `request::urgent`, and others                       |

**Signals kept synchronous (by design):**

| Signal              | Why                                                          |
| ------------------- | ------------------------------------------------------------ |
| `request::manage`   | Lua must apply rules before the client is visible            |
| `request::unmanage` | Client properties must still be valid during the handler    |
| `scanning`/`scanned` | Startup synchronization                                     |

> No rc.lua refactoring required. The observable Lua API is unchanged - handlers still fire, with the same arguments, before the next render.

### What I measured

Ran a controlled A/B between `main` and this branch on non-ASAN bench builds, with rc.lua pinned to the tracked `somewmrc.lua` on both sides and a 15-second scripted workload on a live session.

**The good:**

- `refresh_p99_us` **-23%** (4447 → 3441 µs), `lua_refresh.p99_us` **-22%**, `stages.client.p99_us` **-23%**. Tail latency wins across the per-frame stages, which is what users actually perceive as "smooth."
- `crossings_per_frame.max` **-35%** (31 → 20). Worst-case C↔Lua crossings in a frame collapse as designed.
- `refresh_avg_us` -7% on average too.
- Workload matched: frame counts and signal-emit counts within 1.4% between branches, so both sides are doing the same work.

**The not-so-good (investigated, not blocking):**

- `stages.banning` avg cost went up 4-5x on event-queue. Absolute is still sub-60 µs p99 so the user-time impact is tiny, but the tag-switch client-visibility pass iterates all clients and likely wants a per-tag scene subtree refactor eventually, independent of this queue.
- A pre-GC `lua_memory_kb` snapshot showed +145% on eq. Forcing a full `collectgarbage("collect")` on both branches proved it's transient: Lua registry entries stable at 179 on every state (no `luaL_ref` leak), post-GC residency within 6% of main. It's allocation pressure from per-signal args tables in `event_queue.c`, fully collectable. Poolable as a followup optimization.


JimmyCozza:
I've been daily driving this for a couple weeks.  I think it's a win, certainly easier model for me to wrap my head around.  However, if there are significant issues with this, we can revert it.

Issue #371, opened 2026-03-22
Windows fill workarea when opened in floating layout (regression from #321)


raven2cz:
## Problem

When a tag uses `awful.layout.suit.floating`, newly opened clients fill the entire workarea instead of using their natural/preferred size. This does not happen with tiling layouts.

## Root cause

Commit `acd0fa4` (PR #321, fix for Firefox saved geometry) changed `initialcommitnotify()` to send `wlr_xdg_toplevel_set_size(workarea_width, workarea_height)` for all managed non-float-type clients. PR #334 added a `client_is_float_type()` guard for dialog-type windows, but this doesn't check the tag's active layout.

When the layout is `floating`:
1. `initialcommitnotify()` sends workarea-sized `set_size` → client opens at full workarea size
2. `floating.arrange()` is a no-op → no resize happens
3. Client stays at full workarea size

Before #321, `set_size(0, 0)` was sent, letting clients choose their own size.

## Steps to reproduce

1. Set a tag to `awful.layout.suit.floating`
2. Open any window (e.g. `alacritty`, `ghostty`)
3. Window fills the entire workarea instead of opening at its natural size

## Fix

Use `wlr_xdg_toplevel_set_bounds()` to communicate available space and `set_size(0, 0)` to let clients choose their own size. The bounds hint is sufficient for Firefox's tiling geometry fix — the client knows the maximum available space without being forced into it.

```c
// initialcommitnotify()
if (m && !client_is_unmanaged(c)) {
    wlr_xdg_toplevel_set_bounds(c->surface.xdg->toplevel,
        m->w.width, m->w.height);
}
wlr_xdg_toplevel_set_size(c->surface.xdg->toplevel, 0, 0);
```

Fix branch: [`raven2cz/somewm:fix/floating-layout-initial-size`](https://github.com/raven2cz/somewm/tree/fix/floating-layout-initial-size) (commit `bbd1f97`)

PR #452, opened 2026-04-13
fix: floating layout set_bounds + layer surface send_leave

JimmyCozza:
## Description

Two unrelated C fixes bundled because the first was blocking `main`'s
test suite and I found it while porting the second.

**1. `fix(xdg)`: restore `set_bounds` hint for initial configure.**
The fix from #382 for issue #371 was dropped when `initialcommitnotify`
was extracted into `window.c` in `02528d340`. `tests/test-floating-layout.lua`
was failing on main as a result. This commit restores the
`set_bounds` + `set_size(0, 0)` pattern in its new home.

**2. `fix`: pair `send_leave` with `send_enter` for layer surfaces.**
Forward-port of #450 from `release/1.4` to `main`. On `main` the layer
surface handlers were moved from `somewm.c` to `protocols.c`, so a
straight cherry-pick doesn't apply. Credit for the original diagnosis
and fix to @shuber2. Without this, unplugging a monitor while a
layer-shell client (waybar, fuzzel, tofi, swaync, ...) is running trips
the \`wl_list_empty(&output->events.bind.listener_list)\` assert in
\`wlr_output_finish\`.

## Test Plan

- \`make test-unit\` - 740 passes.
- \`make test-integration\` - 112 passes (floating layout test now green).
- Manual: on a multi-output setup, unplug the secondary monitor with a
  panel running; no assert.

## Checklist

- [x] Lua libraries (\`lua/awful/\`, \`lua/gears/\`, \`lua/wibox/\`, \`lua/naughty/\`) are **not modified** - if a bug surfaces in Lua, the fix belongs in C
- [x] Tests pass (\`make test-unit && make test-integration\`)

### DIFF
--- a/protocols.c
+++ b/protocols.c
@@ -301,6 +301,9 @@ unmaplayersurfacenotify(struct wl_listener *listener, void *data)
 	if (l == exclusive_focus)
 		exclusive_focus = NULL;
 
+	if (l->layer_surface->output)
+		wlr_surface_send_leave(l->layer_surface->surface, l->layer_surface->output);
+
 	if (l->layer_surface->output && (l->mon = l->layer_surface->output->data))
 		arrangelayers(l->mon);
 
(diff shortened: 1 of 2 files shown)


shuber2:
Got a race condition :-)

In #453 I references the original commit in the commit message, which might be helpful. And I added a second commit the improves the code logic a bit.

Issue #446, opened 2026-04-11
1.4.x: Crash after closing firefox tab


shuber2:
## Description

Once in a while, somewm crashes after closing a firefox tab, maybe coinciding with playing videos in the tab. Difficult to reproduce, though.

## Version Info

```
somewm 1.4.0
```

shuber2:
Just trapped into the bug. On dmesg I got:

```
somewm[274511]: segfault at a9 ip 00007fe5aec3a1ca sp 00007fff78480e28 error 4 in libwlroots-0.19.so[721ca,7fe5aebe7000+9d000] likely on CPU 4 (core 2, socket 0)
```

webdesus:
I had this happen when I was trying to open a VLC file that was playing. You can hide VLC and then open it from the tray. It usually opens fine, but once, everything crashed. Looking at the logs, I thought the only Wayland compositor die. I decided it was more of a config issue because I hadn't fully migrated to somewm at that moment yet. 

If looking logs right now i only find this lines:
```
flameshot: The Wayland connection broke. Did the Wayland compositor die?
xdg-desktop-portal-gtk: Error reading events from display: Broken pipe
```
But withot details.

shuber2:
> flameshot: The Wayland connection broke. Did the Wayland compositor die?
> xdg-desktop-portal-gtk: Error reading events from display: Broken pipe

This is the usual consequence of somewm having died. Your `dmesg` or `journalctl` logs should show why somewm died.

webdesus:
Unfortunately, `dmesg` only shows the log for the current session, and I’ve already shut down the computer. `journalctl` doesn’t provide any details either. it just shows that all applications lost the Wayland compositor. I’ll try to look into it more closely next time, but the last time this happened was on April 9.

and i tried look log using next command: 

```
journalctl  --since "2026-04-09 21:25:30" --until "2026-04-09 21:26:30"
```

unai-ndz:
You can use `journalctl -k` as an alternative for dmesg. And `journalctl -k -b -1` is very useful to check the last boot's log.

shuber2:
I just took few minutes and did some investigation based on `segfault at a9 ip 00007fe5aec3a1ca sp 00007fff78480e28 error 4 in libwlroots-0.19.so[721ca,7fe5aebe7000+9d000]`:

```
❯ addr2line -f -e /usr/lib64/libwlroots-0.19.so 0x721ca
wlr_scene_node_coords
[...]

PR #462, opened 2026-04-15
Fix/somewm firefoxtab close crash

shuber2:
## Description

Fixe a use-after-free of scene node accessed via surface->data, but was already destroyed, due to inconsistent and missing cleanup.

## Test Plan

Very difficult. Found in an attempt to tackle issue #446. The reasoning with a canary pointer value is explained in the last commit.

## Checklist

### DIFF
--- a/client.h
+++ b/client.h
@@ -23,16 +23,24 @@ client_is_x11(Client *c)
 	return 0;
 }
 
+static inline bool
+client_has_surface(Client *c)
+{
+#ifdef XWAYLAND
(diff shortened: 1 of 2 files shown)


shuber2:
@JimmyCozza Are you looking into migrating to main, or shall I have a look at it?

JimmyCozza:
@shuber2 If you have the time, that would be great.  Thanks.  This has been a really busy week at work and I've been trying to stay focused.

PR #467, opened 2026-04-16
Fix/use after free wlr scene tree 2.x

shuber2:
## Description

Port of 1.4.x fix of issue #446 .


JimmyCozza (approved):
Awesome!  thanks.

PR #479, opened 2026-04-18
Fix some_is_idle_inhibited() exclude-argument mechanism 1.4

shuber2:
## Description
    fix: Broken exclude mechanism for some_is_idle_inhibited() #446
    
    In destroyidleinhibitor() we call some_is_idle_inhibited() via
    some_recompute_idle_inhibit(), while the being destroyed inhibitor is
    still in the list. In this case the scene tree of the tearing down
    inhibitor surface is already invalid. This is handled by an exclude
    surface argument.
    
    However, when emitting the signal "property::idle_inhibited", again
    some_is_idle_inhibited() is called, but this call is ignorant of this
    exclusion. So we need a different mechanism.
    
    Fix this by using inhibitor->surface->mapped to decide whether we check
    this inhibitor in some_is_idle_inhibited() instead of the exclude
    argument. Remove the exclude argument for some_is_idle_inhibited() and
    some_recompute_idle_inhibit().

## Checklist

### DIFF
--- a/luaa.c
+++ b/luaa.c
@@ -2053,7 +2053,7 @@ luaA_awesome_index(lua_State *L)
 	}
 
 	if (A_STREQ(key, "idle_inhibited")) {
-		lua_pushboolean(L, some_is_idle_inhibited(NULL) || lua_idle_inhibited);
+		lua_pushboolean(L, some_is_idle_inhibited() || lua_idle_inhibited);
 		return 1;
 	}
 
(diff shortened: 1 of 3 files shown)


shuber2:
Hm. I can still trigger the bug, but now very deterministically. Let me investigate further on this PR.

shuber2:
@JimmyCozza This was some annoying ride. But now I think I got it.

shuber2:
@JimmyCozza  Thank you for merging. Are you planning to pull this over to `main` or shall I give it a shot?

Issue #460, opened 2026-04-15
screen.primary setter


unai-ndz:
### What feature would you like?

somewm already inherits the concept of a primary screen from awesome/X11 and uses it in things like `awful.widget.only_on_screen` but as setting that was handled by X11 xrandr there's no api for setting it in somewm.

### Use case

Proper control of `awful.widget.only_on_screen` and similar. Some widgets do not need to be on every screen but right now which screen they show at is "random".

### Proposed API or rc.lua shape

```lua
screen.primary = screen[1]  -- Keeps the getter and setter apis the same and is consistent with awesomewm.

-- Make output store the primary screen instead and delegate screen.primary to it.
-- This is already done for screen.scale but does it make sense here? 
output.primary = output[1]


-- If I remember correctly when I was new to awesomewm I tripped a few times with not being able to do. It could be implemented as QOL.
print(screen[1].primary)
screen[1].primary = true
```

### Prior art

AwesomeWM does through xrandr.
Kwin does through [kscreen-doctor](https://linuxcommandlibrary.com/man/kscreen-doctor)
Gnome Mutter through [gdctl](https://man.archlinux.org/man/gdctl.1.en)

### Preflight checklist

- [x] I have searched existing issues and this feature is not already requested.
- [x] This is a new feature, not a bug fix or an AwesomeWM 4.4 parity gap.

NilOmniscient:
Wait, I thought doing something like this worked fine?

```
local primary_output = output.get_by_name("DP-1")
for s in screen do
  if s.output == primary_output then
    screen.primary = s
  end
end
```

It's what I'm doing in my config to tell the lockscreen API which screen to show the UI on.

unai-ndz:
I may not have tested as thoroughly as I thought I did. I can make it work somewhat but it's pretty unreliable.

I could have swore this returned the same primary all the time but it's working right now.
```lua
screen.connect_signal("primary_changed", function()
    print("PRIMARY SCREEN SIGNAL CHANGED to: " .. (screen.primary.output and screen.primary.output.name or 'nil'))
end)


print("PRIMARY SCREEN IS: " .. (screen.primary.output and screen.primary.output.name or 'nil'))
for s in screen do
    if s ~= screen.primary then
        print("SET SCREEN")
        screen.primary = s
        break
    end
end
print("PRIMARY SCREEN IS: " .. (screen.primary.output and screen.primary.output.name or 'nil'))
```

[...]

PR #481, opened 2026-04-19
fix: set screen.primary in c not in userspace

unai-ndz:
## Description
See #460, although I now think this is more of a bug fix than a new feature.

## Test Plan
```lua
screen.test = 'Dynamic properties work'
print(screen.test)

screen.connect_signal("primary_changed", function()
    print("primary screen signal")
end)

print("primary screen: " .. (screen.primary.output and screen.primary.output.name or 'nil'))

for i = screen.count(), 1, -1 do
    local s = screen[i]
    if s ~= screen.primary then
        screen.primary = s
        break
    end
end
print("new primary screen: " .. (screen.primary.output and screen.primary.output.name or 'nil'))
```

## Checklist
- [ ] Tests pass (`make test-unit && make test-integration`) They are very flaky for me with or without this patch. 

I now think this should be applied to 1.4 as well. I can make a PR if you agree.

### DIFF
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -1850,6 +1850,29 @@ luaA_screen_module_index(lua_State *L)
 	return 1;
 }
 
+/** __newindex metamethod for screen class table
+ */
+static int
+luaA_screen_module_newindex(lua_State *L)
+{
+	const char *key = luaL_checkstring(L, 2);
+
+	/* The .primary property is tracked separately */
+	if (strcmp(key, "primary") == 0) {
+		screen_t *new_primary_screen = luaA_checkscreen(L, 3);
+
+		if ((new_primary_screen) && (new_primary_screen != primary_screen)) {
+			primary_screen = new_primary_screen;
+			luaA_screen_emit_primary_changed(L, primary_screen);
+		}
+		return 0;
+	}
+
+	/* Fallback: Allow arbitrary properties. */
+	lua_rawset(L, 1);
+	return 0;
+}
+
 /** __call metamethod for screen class
  * This function serves as both:
  * 1. The iterator function for "for s in screen do" loops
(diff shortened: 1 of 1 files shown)


JimmyCozza (changes requested):
Thanks @unai-ndz !  

Looks pretty good, just one thing I am trying to think through


### The arbitrary-property fallback.  

If I understand it right, the screen global is a plain table (created by lua_newtable inside luaA_registerlib), not a userdata, and we support LuaJIT through Lua 5.4. Walking through what luaA_getuservalue(L, 1) does on a table for each version, I keep landing on "this might not work":

- LuaJIT / 5.1: wraps lua_getfenv, which pushes nil for a table, then lua_rawset on nil looks like an api_check abort in debug / UB in release.
- 5.2 / 5.3: lua_getuservalue api_checks for ttisuserdata, so a table should abort in debug / read the wrong union field in release.
- 5.4: lua_getuservalue is a macro around lua_getiuservalue with the same userdata precondition.

Even if the write did land, luaA_screen_module_index only does lua_rawget(L, 1) on the screen table itself at objects/screen.c:1847, it never consults a uservalue so reads would
still come back nil. Which makes me think the "dynamic properties" behavior in your test plan probably isn't doing what the comment suggests. But I'd love a sanity check here in case I'm missing something about how the module table is set up.

unai-ndz:
You are right indeed. My bad. I tested a few things and ended up mixing up the commits. With the last modification it does pass the test plan.
Also thanks for the detailed comment, it is very helpful.

JimmyCozza (approved):
Thanks!  Looks great

shuber2:
@JimmyCozza This never got merged to release/1.4, right? It would replicate the X11 feature `xrandr --output foo --primary` based on which we could have different wibars for different screens as in awesomewm.

PR #499, opened 2026-04-22
fix: Broken exclude mechanism for some_is_idle_inhibited() #446

shuber2:
In destroyidleinhibitor() we call some_is_idle_inhibited() via some_recompute_idle_inhibit(), while the being destroyed inhibitor is still in the list. In this case the scene tree of the tearing down inhibitor surface is already invalid. This is handled by an exclude surface argument.

However, when emitting the signal "property::idle_inhibited", again some_is_idle_inhibited() is called, but this call is ignorant of this exclusion. So we need a different mechanism.

Fix this by using inhibitor->surface->mapped to decide whether we check this inhibitor in some_is_idle_inhibited() instead of the exclude argument. Remove the exclude argument for some_is_idle_inhibited() and some_recompute_idle_inhibit().

Forward-ported from release/1.4 commit 1abe77fe.

## Checklist

### DIFF
--- a/luaa.c
+++ b/luaa.c
@@ -2158,7 +2158,7 @@ luaA_awesome_index(lua_State *L)
 	}
 
 	if (A_STREQ(key, "idle_inhibited")) {
-		lua_pushboolean(L, some_is_idle_inhibited(NULL) || lua_idle_inhibited);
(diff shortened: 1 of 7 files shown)


JimmyCozza (approved):
👍

PR #512, opened 2026-04-25
fix: Sync meson license with actual LICENSE file

shuber2:
## Description
Meson build file had MIT license set, where the LICENSE file is GPL-3. Applies to main and release/1.4.


JimmyCozza:
Yikes!  Thank you!

PR #522, opened 2026-04-28
fix(mouse): defer aerosnap placeholder by dwell to kill cross-monitor flicker

raven2cz:
## Description

Dragging a floating client across the seam between two outputs on a Wayland scene-graph compositor produces a visible single-frame placeholder flicker on the destination monitor. `detect_areasnap()` shows the aerosnap placeholder on the very first tick the cursor enters an edge band, even when the cursor is just transiting and parks somewhere far from the edge.

This patch debounces the placeholder by `awful.mouse.snap.snap_dwell_ms` (default 150 ms): the placeholder is only shown once the cursor has dwelled in the same edge zone for that long. Fast cross-monitor crossings graze the edge band for under 50 ms and so never trigger the placeholder; the destination output is never damaged.

### Behaviour

| Scenario | Before | After |
|----------|--------|-------|
| Cross-monitor drag (cursor flies past edge) | Placeholder flashes on destination output for one frame | No placeholder, no damage |
| Park cursor near edge for >150 ms | Placeholder appears on first edge-detect tick | Placeholder appears after 150 ms |
| Release inside edge zone with no prior placeholder | Snap applies | No-op (no visual confirmation, no commit) |
| `module.snap_dwell_ms = 0` | n/a | Legacy first-tick behaviour preserved |

### State changes

The dwell mechanism required tightening surrounding state: a `current_snap_visible` flag tracks whether `show_placeholder` was ever called for the current zone (gates the release decision on actual visual feedback rather than just elapsed time), and after a successful apply the snap/axis are snapshotted to locals before `reset_snap_state()` clears all globals so an immediate same-edge follow-up isn't blocked by stale state. A small `show_snap_for(c)` helper deduplicates the `awful.placement` call between the two code paths.

### Test plan

- [x] Drag floating windows between all three pairs of monitors at varying speeds: placeholder no longer flashes on transit
- [x] Park cursor at left edge for ~200 ms: placeholder appears, snap commits on release
- [x] Park cursor at left edge briefly (<100 ms), release: no surprise snap
- [x] Drag, snap left, drag again, snap left: second snap works (regression check)
- [x] `awful.mouse.snap.snap_dwell_ms = 0` at runtime: legacy behaviour restored
- [x] `luac -p lua/awful/mouse/snap.lua` clean

Originally hit on a multi-monitor NVIDIA setup where per-frame placeholder + cross-output damage was particularly visible, but the fix is platform-agnostic.

### DIFF
--- a/lua/awful/mouse/snap.lua
+++ b/lua/awful/mouse/snap.lua
@@ -14,6 +14,7 @@ local beautiful = require("beautiful")
 local color     = require("gears.color")
 local shape     = require("gears.shape")
 local cairo     = require("lgi").cairo
+local GLib      = require("lgi").GLib
 local alayout   = require("awful.layout")
 
 local capi = {
@@ -26,7 +27,13 @@ local capi = {
 
 local module = {
     default_distance  = 8,
-    aerosnap_distance = 16
+    aerosnap_distance = 16,
+    -- Minimum time the cursor must dwell inside an aerosnap edge zone
+    -- before the placeholder is shown.  Flying past an edge during a
+    -- cross-monitor drag (typical: <50 ms in the edge band) never
+    -- lights up the placeholder and never damages the second output.
+    -- Set to 0 to restore the legacy behaviour (show on first tick).
+    snap_dwell_ms     = 150,
 }
 
 local placeholder_w = nil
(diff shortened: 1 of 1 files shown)


JimmyCozza:
thanks @raven2cz

PR #552, opened 2026-05-09
fix(drawin): assign fake-screen wibars via Lua screen registry

JimmyCozza:
## Description
`drawin_assign_screen` only consulted `some_monitor_at`, which knows
wlroots monitors only. Wibars attached to fake screens (created via
`screen.fake_add()`) therefore got attributed to the focused real
monitor instead of their fake one. `screen_update_workarea` then
applied the strut to the wrong screen, leaving the fake screen's
workarea equal to its full geometry.

Fall back to `screen_getbycoord()` (which iterates the Lua screen
registry, including fake screens) before the focused-monitor fallback.
Real-screen behavior is unchanged because `some_monitor_at()` still
takes precedence whenever it returns a result.

## Test Plan
- New regression test `tests/test-multi-screen-wibar-workarea.lua`:
  creates a fake screen via `screen.fake_add`, attaches a wibar to the
  primary and the fake screen, asserts both workareas shrink by the
  wibar height. Passes.
- `make test-unit`: 758/758 pass.

## Checklist

### DIFF
--- a/objects/drawin.c
+++ b/objects/drawin.c
@@ -691,20 +691,24 @@ static void
 drawin_assign_screen(lua_State *L, drawin_t *drawin, int drawin_idx)
 {
 	Monitor *m;
-	screen_t *new_screen;
+	screen_t *new_screen = NULL;
 	screen_t *old_screen = drawin->screen;
 
-	/* Get monitor at drawin's position */
+	/* Prefer the wlroots layout: it covers physical outputs accurately. */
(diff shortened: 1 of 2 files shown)


JimmyCozza:
Closing. This only addresses a narrow fake-screen variant and doesn't fix the original real-screen-2 startup bug it was investigated for. Revisiting once the real bug is root-caused.

PR #597, opened 2026-06-01
fix(nix-build): correct lgi-glib-2.88 patch to replace get() with ref()

### LINKED ISSUE #594: Nix build crashes on lgi glib 2.88 enum patch
### Branch

main (2.0-dev, unstable)

### somewm --version output

```shell
## somewm version info

**somewm:** 2.0.0-dev (unknown)
**wlroots:** 0.19
**Lua:** LuaJIT 2.1.1741730670 (compiled: Lua 5.1.4)
**LGI:** 0.9.2
**Build:** D-Bus=yes, XWayland=yes

**System:**
- Distro: NixOS 26.05 (Yarara)
- Kernel: 6.18.31
- Arch: x86_64
- GPU: nvidia (10DE:1C8D)
- Session: wayland (nested: yes)
```

### How did you install somewm?

Distro package (AUR, nixpkgs, etc.)

### What happened?

Tried to update somewm via nixpkgs

### Steps to reproduce

1. Use nix
2. Install/update to latest somewm version

### Expected behavior

_No response_

### Relevant rc.lua

```lua

```

### Logs, crash output, or backtrace

```shell
┃        Last 13 log lines:
┃        > structuredAttrs is enabled
┃        > Running phase: unpackPhase
┃        > unpacking source archive /nix/store/xpfpa47gfiw8rlz0fkcf4mwyyq89gl47-source
┃        > source root is source
┃        > Running phase: patchPhase
┃        > applying patch /nix/store/z8ish0nvw71dzf9cwla35pswjg6a6aws-lgi-find-cairo-through-typelib.patch
┃        > patching file lgi/override/cairo.lua
┃        > applying patch /nix/store/6wz23hgrhkwpg9hp8pp2yx0n1p91d9l8-glib-2.86.0.patch
┃        > patching file lgi/ffi.lua
┃        > applying patch /nix/store/fjk1ngd7w1bgq819ny1r0mlb02iagv6r-lgi-glib-2.88-enum.patch
┃        > patching file lgi/ffi.lua
┃        > Hunk #1 FAILED at 75.
┃        > 1 out of 1 hunk FAILED -- saving rejects to file lgi/ffi.lua.rej
```

### Preflight checklist

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have tested with a minimal `rc.lua` (or `rc_fallback.lua`) to confirm this is not a bug in my own config.
- [x] I have read [CONTRIBUTING.md#branches](https://github.com/trip-zip/somewm/blob/main/CONTRIBUTING.md#branches) and selected the correct branch above.

se-neaxplore:
The patch was missing the `-` marker, leaving the get→ref substitution as a context line instead of an actual replacement.

Fixes: #594


JimmyCozza (changes requested):
Thanks @se-neaxplore.  From what I can tell, this works, but I don't think we should land it.  I don't think we need to carry `lgi-glib-2.88-enum.patch` at all anymore (or if we ever did tbh).

somewm builds lgi from nixpkgs, and nixpkgs already patches lgi for the GLib 2.88 enum change itself - `NixOS/nixpkgs@79f6dec`, "Backport fix for GLib 2.88" (2026-05-25). Arch does the same in their lgi PKGBUILD (`0001-Use-TypeClass.get-instead-of-.ref.patch` +  `0002-ffi-conform-load_enum-to-GLib-2.87.0.patch`). Both carry it on lgi 0.9.2 themselves, since upstream hasn't merged a fix, so our patch is stacked on an lgi that's already fixed.

  - https://github.com/NixOS/nixpkgs/commit/79f6dec101e0ed82160022dde77b88c52c0ff892
  - https://gitlab.archlinux.org/archlinux/packaging/packages/lgi/-/blob/main/PKGBUILD

I'd drop `lgi-glib-2.88-enum.patch` and the `lgi.overrideAttrs` block in `package.nix` and just use nixpkgs' lgi. Would you be up for repurposing this PR to do that? Happy to help if anything comes up.

JimmyCozza (approved):
Image

It's so beautiful

PR #605, opened 2026-06-04
Feat/clay everywhere

JimmyCozza:
_Oops: accidentally merged the original (#527) into main and reverted it (#604). This is the replacement tracking PR; original description below._

---

 ***This branch is an experiment.***

  ### Background

  I am not a particularly smart man. I frequently get confused, especially trying to keep this whole codebase and my whole codebase at work in my head.

  One of the frequent places I get lost is in how the layouts are defined for the distinct elements in the compositor.
  - What's the underlying algorithm that defines the geometry of the wibar(s)?
  - How about clients in layout x? 
  - What about a custom widget I've created inside a titlebar with 3 sibling elements?

  Then I started thinking, "Why are they using different rules for placement when unified layout libraries exist?"

  [Clay](https://www.nicbarker.com/clay) is a single-header C layout engine by Nic Barker that feels like it fits what I want out of a "unified" layout approach. C native, simple to understand, built-in and well documented layout rules, reasonable performance (certainly up for debate, but IMO it's worth the simpler mental model).

  The old architecture has many layout helpers (each tag layout, each wibox container, each placement function) doing their own pixel arithmetic in Lua and pushing geometry into C. The C side ends up as a fan-in of many parallel call sites that all want to do the same thing. The hope is that Clay collapses those into one solver entry point that Lua orchestrates and C executes.

  ### Before: how a layout resolves on AwesomeWM / somewm 1.4

  There is no single flow; each domain has its own.

  **Tag layout** (clients on a tag):
  1. `tag::arrange` fires; the current layout's `arrange(p)` function runs.
  2. The module does its own loop: master/slave split, gaps, per-client rects.
  3. Calls `c:geometry({...})` per client as it goes.
  4. C writes the geometry, configures the xdg surface, moves the scene node; wlroots renders.

  **Wibox widgets**:
  1. Wibox needs redraw; `wibox.hierarchy` walks the widget tree.
  2. Each `wibox.layout.*` / `wibox.container.*` has its own `:layout(context, w, h)` method returning child placements.
  3. Hierarchy composites the placements; cairo draws; result becomes a `wlr_scene_buffer`; wlroots renders.

  **Anchored placement** (`awful.placement.top_right`, etc.):
  1. The helper reads screen geometry and computes one rect imperatively.
  2. `no_overlap` and `no_offscreen` are separate functions that loop obstacles or clamp edges.
  3. Calls `:geometry()`; C drives wlroots.

  **Wibar**:
  1. Constructed; a deferred timer fires on the next frame.
  2. Reads screen geometry, does its own anchoring math, sets struts.
  3. Sets drawin geometry; C drives wlroots.

  These four paths share only the geometry write and the workarea read. The layout algorithms themselves are duplicated.

  ### After: how a layout resolves on this branch

  1. **Lua builds a descriptor tree** for whatever needs placing (`{_type, props, children}`): a tag's clients, a wibox's widgets, an anchored popup. Built at the moment the object is created or its inputs change (no deferred timer).
  2. **`somewm.layout.solve(descriptor)`** is the single entry point.
  3. **Clay (in C) computes the rects** from the tree: workarea, gaps, splits, anchoring, avoid/clamp, decoration math.
  4. **Solver returns** `{placements = {[object] = {x, y, w, h}}, workarea = {...}}`.
  5. **Lua applies the rects** by calling `:geometry({...})` on each object.
  6. **C drives wlroots**: `:geometry` writes to the client/drawin struct, configures the xdg surface, and moves the scene graph node.
  7. **wlroots renders** the scene on the next frame.

  The shape is the same for tag layouts, wibox widgets, placement, and the wibar; only the descriptor type and the kind of object being positioned differ.

  ### Why this branch/PR might not stick

  - Per-call `c:geometry({...})` cost went from 0.49ms to 1.13ms (~2.3x). Fine for single calls at 60Hz; batched paths could feel it.
  - A handful of layouts (magnifier, mouse-anchored placement, popup terminal anchor, `naughty.layout.box`) did not fit the descriptor model cleanly and are still on the old path. If I find that this list keeps growing while I am daily driving this branch, the abstraction is either completely wrong or I need to tweak it.
  - `DEVIATIONS.md` lists 9 placement/layout carve-outs that still bypass Clay. I want that list to shrink during the week, not stay flat.

  ### What's in the code

  1. Clay engine in C: per-screen contexts, gap handling, decoration sub-pass, layer-shell anchoring, hot-reload safety
  2. `awful.layout.suit.*` ported to `awful.layout.clay.*`; old suit modules alias or are removed
  3. `somewm.layout` unified solver (with a pure-Lua reference solver for the busted environment); all wibox layouts/containers migrated
  4. Anchored placement via `somewm.placement.solve`, `no_overlap` via Clay avoid, `no_offscreen` via Clay clamp, synchronous wibar compose, `DEVIATIONS.md`

  ### Final thoughts

  I might end up just completely abandoning this if it doesn't fit, if it changes the public lua API too much, if it's too slow performance-wise as I write more and more nested widgets, or anything else. If it's not good enough, I'll just happily abandon it and move back to more normal dev work.

  There's 1 more piece of this that's not technically related, but understanding the full layout and rendering cycle has put me in a better position where I could test out what it might look like to try a GPU only rendering approach to all compositor level elements. I'm not quite saying that this PR is needed for that, but this helped me wrap my head around what it would take to funnel all of the layout compute step through a single path (clay) and I think it could be valuable to funnel all of the rendering step through a single path.  

Anyway, try this out, don't, whatever.  I just wanted to commit some of the stuff I've been busy with over the last week or so so I can pause and get back to fixing the issues that have been piling up while I've been heads down.


JimmyCozza:
OK, the really short version of this (And I should write up something more long form) is that I think clay is a really great choice for this concept and I want to use it.

However, I'm not satisfied with this implementation.  It was good for validating that clay is a solid choice for all things layout, and even more exciting it comes with a couple things for free.

1) A debug inspector built-in.  You can hover over the inspector and get highlighting of the widget as if it were a browser hovering over css declarations in the inspector.  
2) 1 unified tree per screen.  You can print out the tree at any moment and get every element dumped for free.
3) Completely offloads all geometry, layout, and other positional work from wibox and awful.layout.  lua would only need to describe the widget and where it fits in the tree.
4) This would basically make it completely possible to take the clay layout array and pass it into a rendering function.  So hypothetically, we could test out GPU based rendering much more easily down the road.

Anyway, this branch was cool, educational, but not worthy of merging.

Issue #360, opened 2026-03-19
Carousel layout: followup enhancements from PR #351


JimmyCozza:
During testing of PR #351, @oeleonor and @vredesbyyrd suggested several enhancements beyond the bugs fixed for merge. Tracking them here as followups.

- [ ] **Reorder clients within a column** (@oeleonor) - Add an API to rearrange stacked clients within a column. `push_window` could evolve for this, similar to the `bydirection` client functions.
- [ ] **Don't clip partially-visible clients** (@oeleonor) - Clients still partially on screen should not have their scene nodes clipped. Noted as personal preference.
- [ ] **Dynamic peek width** (@vredesbyyrd) - Disable peek on sides where no offscreen columns exist, so there is no empty space at edges. niri has the same issue; vredesbyyrd worked around it with IPC + conditional struts.
- [ ] **Per-client column width via rules** (@vredesbyyrd) - Override `default_column_width` per client using `ruled.client` properties (e.g. `carousel_column_width = 0.75`). Fits naturally into AwesomeWM's existing rule system.
- [ ] **Viewport info for widgets/IPC** (@vredesbyyrd) - Expose scroll position, strip dimensions, and column positions so widgets can build a "position in viewport" indicator.

### Bugs
- [ ] Floating focus shifts viewport
- [ ] Vertical gaps halved vs horizontal gaps
- [ ] Fullscreen round-trip moves client to end
- [ ] Tag toggle order depends on focus

oeleonor:
In #642 I've implemented most features/fixes listed here. From the original list:

Features:
- [x]  Reorder clients within a column
- [x]  Don't clip partially-visible clients (Address by #491/#492)
- [x]  Dynamic peek width
- [x]  Per-client column width via rules
- [x]  Viewport info for widgets/IPC

Bugs:
- [x]  Floating focus shifts viewport
- [x]  Vertical gaps halved vs horizontal gaps
- [ ]  Fullscreen round-trip moves client to end; due to columns only containing tiled clients, also applies to floating clients and presumably any other client state that is mutually exclusive with tiled
- [ ]  Tag toggle order depends on focus; due to no stack shared across the screen, but multiple stacks for each tag

I'll be taking a look at column width via rules later this week, but I don't think I have much desire to work on the 2 remaining bugs (especially the fullscreen one).

JimmyCozza:
I'm going to close this ticket.  Those 2 remaining bugs are rendered obsolete in the new code I've got on 2.1.

[...]

PR #642, opened 2026-07-13
Feat/carousel followup enhancements

oeleonor:
## Description
Addresses some of the behaviors described in #360. Notably:
Adds `carusel.push_window_bydirection()`, which can be used to reorder clients within a column and includes `carousel.push_window()` column boundary behavior.
Floating focus no longer shifts the viewport,
Vertical gaps now match horizontal gaps,
Adds dynamic peeking, enabled via `beautiful.carousel_dynamic_peek_width`, which currently removes the empty peek space from edge columns, but does not resize any columns,
Adds `carousel.get()`, which provides layout information for widgets/IPC,
Adds `client.carousel_column_width` property definable via ruled, which specifies client-specific default column widths.
Reverts regression on carousel.lua from a7ecae7

## Test Plan
Daily usage

### DIFF
--- a/lua/awful/layout/suit/carousel.lua
+++ b/lua/awful/layout/suit/carousel.lua
@@ -20,6 +20,7 @@
 local capi = { client = client, screen = screen, awesome = awesome }
 local math = math
 local ascreen = require("awful.screen")
+local prior_ci
 
 -- awful.layout is lazy-loaded: carousel is required during awful.layout init,
 -- so a top-level require would be circular. We cache after first use.
@@ -404,8 +405,9 @@ function carousel._arrange_impl(p, vertical)
 
     -- Compute target scroll offset based on centering mode
     local focus_ci = focused_col_idx(state, focus)
-    if not focus_ci then focus_ci = 1 end
+    if not focus_ci then focus_ci = prior_ci - 1 end
     state.last_focused_ci = focus_ci
+    prior_ci = focus_ci
 
     local center_mode = get_beautiful().carousel_center_mode or carousel.center_mode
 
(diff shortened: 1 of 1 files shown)


oeleonor:
This also addresses double animations during certain actions. Fixed by removing carousel-specific animation timer while global animations are enabled, I imagine the desired fix would be to remove the carousel-specific animation code entirely to defer to global animations.

JimmyCozza (dismissed):
looks great, thanks @oeleonor 

Couple things

- push_window no longer accepts the old numeric args (-1/1), they fall through to a silent no-op. Breaks test-carousel-push-window.lua and existing keybindings.  
- focus_ci = state.last_focused_ci - 1 can crash: last_focused_ci is nil on floating focus, and the result can exceed #state.columns after coalesced arranges. I think it needs a nil guard and math.min(...,#state.columns).
- carousel.get() throws on a tag that's never been arranged (state.workarea nil, dp >= 0 on a boolean). Also focus_ri is always nil, and it leaks the live state.columns table instead of returning plain data.

oeleonor:
These 3 should be fixed now. I'm noticing an issue with end-of-strip columns using peek_width instead of dynamic_peek_width when the prior end-of-strip column is destroyed. I'd like to fix that before merging.

I'm thinking it would actually be trivial to implement a single column table for the screen rather than for each tag, which would preserve column positions across tags similarly to client positions in typical layouts, but it would cause some awkward behavior with multi-client columns where the clients don't share the same tags. Do you have any thoughts?

oeleonor:
Tracking all the clients attached to the screen works well but feels very un-awesome, so I think I'm done here. I'll keep anything else to a separate pull request as I'll want to order and move clients based off p.state instead of an internal table, and it'll take a bit of a rewrite.

JimmyCozza:
Thank you so much!  I'm on the final day of a road-trip.  WHen I get to civilization I'll do 1 more check through it but this looks great!  I appreciate all your work

PR #643, opened 2026-07-14
nix: Fix SVG support

seqizz:
## Description

Very small change, fixes SVG support, did not open my config otherwise.

## Test Plan

Currently trying to migrate

## Checklist

### DIFF
--- a/package.nix
+++ b/package.nix
@@ -28,6 +28,7 @@
   xwayland,
   gtk3Support ? true,
   gtk3 ? null,
+  wrapGAppsHook3,
   extraGIPackages ? [ ],
   extraLuaPackages ? (_: [ ]),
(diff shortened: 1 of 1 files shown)


seqizz:
A bit more info, as far as I could understand gdk-pixbuf had librsvg as a dependency but couldn't find the loader at runtime without the loader cache. wrapGAppsHook3 is doing some magic about GDK_PIXBUF_MODULE_FILE and make it discover.

JimmyCozza (approved):
Thanks @seqizz 

I did some pretty rudimentary testing on this.  It worked OK on a nix machine I have, though I'm not a fulltime nix user, so I am mostly just assuming that this is idiomatic.

Issue #626, opened 2026-07-05
beautiful.border_width introduces unwanted bottom+right border


potamides:
### Branch

release/1.4 (stable, AwesomeWM 4.4 parity)

### somewm --version output

```shell
## somewm version info

**somewm:** 1.4.2 (2026-06-22)
**wlroots:** 0.19
**Lua:** LuaJIT 2.1.1780076327 (compiled: Lua 5.1.4)
**LGI:** 0.9.2
**Build:** D-Bus=yes, XWayland=yes

**System:**
- Distro: Arch Linux
- Kernel: 7.0.11-arch1-1
- Arch: x86_64
- GPU: nvidia (10DE:28BA)
- Session: wayland (nested: yes)
```

### How did you install somewm?

Distro package (AUR, nixpkgs, etc.)

### What happened?

First thank you for this project! I am having a lot of fun with porting my config to SomeWM. I noticed that a border_width > 0 introduces a gap at the bottom and right side of a client. Please check this screenshot created with the default config and an oversized border_width to illustrate the problem:

<img width="3440" height="1440" alt="Image" src="https://github.com/user-attachments/assets/46a4ad3e-5089-46d2-856d-5d0433bbb7d1" />

Btw, I just see that the actual border is invisible in the screenshot. That's not how it looked like on the screen and might be an issue with the screenshot tool. In any case the increased border width at the bottom and right side are still clearly visible.

### Steps to reproduce

1. Use tile layout
2. increase border_width

### Relevant rc.lua

```lua

```

### Logs, crash output, or backtrace

```shell

```

### Preflight checklist

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have tested with a minimal `rc.lua` (or `rc_fallback.lua`) to confirm this is not a bug in my own config.
- [x] I have read [CONTRIBUTING.md#branches](https://github.com/trip-zip/somewm/blob/main/CONTRIBUTING.md#branches) and selected the correct branch above.

notFoxils:
I'm also experiencing this on NixOS, so this doesn't seem to be arch specific.

Additionally, I've found a (pretty simple) [fix](https://github.com/notFoxils/somewm/commit/ecbd9338981fcec0b43de4d68874629fa558da14), but it was with a modification to the `awful` library...
Because of that, this patch seems to be a no-go (due to lua lib modification) according to the default description when making a pull request.

[...]

PR #645, opened 2026-07-18
#626 | fix(wlroots-sync): Stop accounting for border width in client geometry

notFoxils:
## Description
- `apply_geometry_to_wlroots`
  Fix border positioning and size calculations.
  Stop subtracting border width from client geometry when setting
  client size.
- `client_get_clip`
  Stop accounting for border width when getting clip dimensions.
- `client_resize`
  Fix the comment :P

## Test Plan
image
Tested with obnoxious border sizes, extra bottom right space is gone.

## Checklist

### DIFF
--- a/DEVIATIONS.md
+++ b/DEVIATIONS.md
@@ -25,8 +25,9 @@ This document tracks all known differences between somewm and AwesomeWM. These e
 
 **Titlebar Border Positioning**
 - In X11, borders are drawn OUTSIDE the window frame by the X server
-- In Wayland, borders are scene rects at geometry edges
-- Titlebars must start INSIDE the border area, hence `border_width` inset
+- In Wayland, somewm draws borders as scene rects around the geometry: the scene tree origin is the outer border corner and the client footprint is `geometry` plus `border_width` on each side
+- `c->geometry` excludes the border, matching AwesomeWM
+- Titlebars are inside the geometry and must start INSIDE the border area, hence `border_width` inset
 - See `titlebar_get_area()` in `objects/client.c`
 
 **WM Restart**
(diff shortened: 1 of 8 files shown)


JimmyCozza (dismissed):
Thanks @notFoxils.  This looks really great.

Requesting changes that weren't your fault, just while you're in here I wonder if you mind: a few other consumers of `c->geometry` still assume the old border-inclusive model and now regress the opposite way for anyone using `border_width`:

  1. **`titlebar_get_area`** (objects/client.c) still does `width -= 2 * bw`, with bottom/right offset by `- bw`. Under the new model that should be dropped, with bottom at `bw + geometry.height -
  size` and right at `bw + geometry.width - size`.
  2. **Shadow sizing** in `apply_geometry_to_wlroots` (window.c, just below your change) passes bare `geometry.width/height` — needs `+ 2 * c->bw` since the frame footprint now includes the border.
  3. **Map handler** (window.c): `c->geometry.width += 2 * c->bw;` (and height) should just be deleted — it inflates new floating clients by 2*bw.


  Also: drop the flake.lock bump

notFoxils:
I re-based onto main and deleted the lock bump in the process.

The requested changes are made, with an exception in the shadow sizing.
If I use `+ 2 * c->bw` instead of `+ c->bw` I end up with an extra pixel of shadow.
So with this configuration I'd end up with 11 blue pixels extended from the right and bottom of the window instead of the expected 10.
```lua
theme.shadow_color = "#0000FF"
theme.shadow_opacity = 1
theme.shadow_radius = 1
theme.shadow_offset_x = 10
theme.shadow_offset_y = 10
theme.shadow_enabled = true
```

JimmyCozza:
Thanks @notFoxils, looks great. I pushed a commit on top covering a few more consumers of the old border-inclusive model, plus a regression test.

PR #646, opened 2026-07-18
fix(screen): C-side `screen:get_bounding_geometry` parity with old lua functionality

notFoxils:
## Description
- `luaA_screen_get_bounding_geometry`
  Apply the margin-based geometry adjustments like was done before
  in the lua function `screen.object.get_bounding_geometry`.

## Test Plan
image
image
image

Tested with obnoxious useless_gap sizes in the theme and checking that the column, gap, and margin sizes are consistent.

## Checklist

## Notes
The function is still missing padding adjustments I want to talk
before I make these changes as I believe it need an additional
field in screen_t and getters/setters for that field.
The documentation above the function probably needs adjustment.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1771848320,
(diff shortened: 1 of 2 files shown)


shuber2:
I guess this then also applies to `release/1.4`?

PR #664, opened 2026-08-06
fix(input): clickfinger_button_map missed NULL init

shuber2:
A minor fix of a missing NULL init of a string variable.


JimmyCozza (approved):
Thanks @shuber2

PR #680, opened 2026-08-09
fix(monitor): detach scene output before Lua removal work in cleanupmon

JimmyCozza:
## Description
Fixes a crash on output unplug. Relocating clients off the removed screen re-sent
surface enter on the dying output while its scene output was still alive, leaving a
bind listener that tripped the `wlr_output_finish` assert. Also adds
`awesome._test_remove_output` and moves the multi-screen tests onto real headless
output hotplug, which is what exposed the crash.

## Test Plan
`make test` passes (unit, check, signal, orchestrator, restart, integration).
The migrated multi-screen tests also pass under `HEADLESS=1`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ All notable changes to somewm will be documented in this file.
 
 ### Added
 
+- `awesome._test_remove_output` test helper - destroys a headless output by name, driving the real unplug path
+
 ### Fixed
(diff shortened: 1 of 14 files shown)


JimmyCozza:
Actually, I have a better idea for how to do this later in 2.2 I think.  Not going to support it immediately.

Issue #663, opened 2026-08-06
feature: Add support for tablet wayland protocol


shuber2:
## Overview

Tablet drawing devices – e.g., Wacom tablets – are supported in X11 via xf86-input-wacom. For wlroots-based wayland compositors, the hardware events are passed via libinput to the wlroots core, then to the compositor and then need to be passed to the client. To this end, somewm needs to implement the *tablet-v2* protocol defined by `stable/tablet/tablet-v2.xml`.

Some Documentation:
* https://linuxwacom.github.io/
* https://deepwiki.com/newm-next/wlroots/5.4-tablet-input
* https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html and in particular `man libinput-debug-tablet`

## Somewm branching model

Besides somewm-2.x, we probably also want this for somewm-1.4, I believe. Though it is not a missing feature from awesomewm, it is a missing feature from awesomewm/xorg in a sense.

Maybe implementing it in release/1.4 first and then porting it to main is a reasonable way, as the port also requires some architectural decisions, e.g., creating `tablet.{c,h}`. Here we should also keep support for touch devices (#612, #524) in mind.

I have not look too deep into this, but maybe an implementation can be done in a two stages:
* A minimal viable version where the pen can be used effectively as pointer.
* A full tablet version that allows clients (krita, xournal++) that are tablet-aware to use pressure or rotation information as intended.

### Use case

* Drawing, like in krita or gimp
* Annotating, like in okular (either in document or presentation mode)
* Writing, like in xournal++

In my opinion, there are two valuable sources of what can possibly be configured:
* `man xsetwacom` from xf86-input-wacom
* `man sway-input`, in particular the `map_to_output` options.

### Proposed API or rc.lua shape

```lua
-- All settings inspired from `man sway-input`
awful.input.rules = {
    -- Settings for all tablets
    { rule = { type = "tablet" },
      properties = {
          map_to_output = "*", -- wildcard means global layout, and is the fallback. can be output name (e.g., DP-8) or a "make model serial" string.
      } },
    -- Tablet-specific settings
    { rule = { name = "Wacom Intuos S Pen" },
      properties = {
          map_to_output = { "DP-8", "AOC U27E3U ZX3QBHA000360" }, -- list of candidates, "*" is the implicit fallback
          -- map_to_region = "100x100 1000x1000", -- not mapping to an output, but to a region
          map_from_region = "X1xY1 X2xY2" -- defaults to "0x0 1x1"
[...]

PR #692, opened 2026-08-16
feat(input): add tablet and touch support for release/1.4

shuber2:
## Description

It adds tablet support as described in feature issue #663 and touch support as described in feature issue #707.

The following developments could naturally follow up on this PR:
 - Adding support for input rule management of somewem-client.

## Test Plan

Only little test code coverage. Mainly manual tests.

## Checklist

### DIFF
--- a/somewm.c
+++ b/somewm.c
@@ -3268,7 +3268,10 @@ inputdevice(struct wl_listener *listener, void *data)
 		createpointer(wlr_pointer_from_input_device(device));
 		break;
 	default:
-		/* TODO handle other input device types */
+		wlr_log(WLR_INFO,
+			"[input] ignoring unsupported device type=%d name=%s",
+			device->type,
+			device->name ? device->name : "(unknown)");
 		break;
 	}
 
--- a/somewm.c
+++ b/somewm.c
@@ -65,6 +65,9 @@
 #include <wlr/types/wlr_session_lock_v1.h>
 #include <wlr/types/wlr_single_pixel_buffer_v1.h>
 #include <wlr/types/wlr_subcompositor.h>
+#include <wlr/types/wlr_tablet_pad.h>
+#include <wlr/types/wlr_tablet_tool.h>
+#include <wlr/types/wlr_tablet_v2.h>
 #include <wlr/types/wlr_viewporter.h>
 #include <wlr/types/wlr_virtual_keyboard_v1.h>
 #include <wlr/types/wlr_virtual_pointer_v1.h>
@@ -147,6 +150,48 @@ typedef struct {
 	struct wl_list link;
 } TrackedPointer;
 
+/* Tracked tablet device for tool-event listeners and debug logging */
+typedef struct {
+	struct wlr_tablet *tablet;
+	struct wlr_tablet_v2_tablet *tablet_v2;
+	struct wl_listener axis;
+	struct wl_listener proximity;
+	struct wl_listener tip;
+	struct wl_listener button;
+	struct wl_listener destroy;
+	struct wl_list link;
+} TrackedTablet;
+
+/* Tracked tablet pad for debug logging in MVP phase */
+typedef struct {
+	struct wlr_tablet_pad *pad;
+	struct wlr_tablet_v2_tablet_pad *pad_v2;
+	struct wlr_tablet *attached_tablet;
+	struct wlr_surface *current_surface;
+	struct wl_listener button;
+	struct wl_listener ring;
+	struct wl_listener strip;
+	struct wl_listener attach_tablet;
+	struct wl_listener surface_destroy;
+	struct wl_listener destroy;
+	struct wl_list link;
+} TrackedTabletPad;
+
+typedef struct {
+	struct wlr_tablet_tool *tool;
+	struct wlr_tablet_v2_tablet_tool *tool_v2;
+	struct wlr_tablet *tablet;
+	/* Last known normalized (0..1) position and tilt, tracked per-tool so
+	 * a single-axis wlr_tablet_tool_axis_event (see tabletnotifyaxis())
+	 * can still resolve a full x/y or tilt_x/tilt_y pair. */
+	double last_x, last_y;
+	bool has_x, has_y;
+	double last_tilt_x, last_tilt_y;
+	bool has_tilt_x, has_tilt_y;
+	struct wl_listener destroy;
+	struct wl_list link;
+} TrackedTabletTool;
+
 /* function declarations */
 static void applybounds(Client *c, struct wlr_box *bbox);
 void arrange(Monitor *m);
@@ -179,6 +224,8 @@ static void createlocksurface(struct wl_listener *listener, void *data);
 static void createmon(struct wl_listener *listener, void *data);
 static void createnotify(struct wl_listener *listener, void *data);
 static void createpointer(struct wlr_pointer *pointer);
+static void createtablet(struct wlr_tablet *tablet);
+static void createtabletpad(struct wlr_tablet_pad *pad);
 static void createpointerconstraint(struct wl_listener *listener, void *data);
 static void createpopup(struct wl_listener *listener, void *data);
 void cursorconstrain(struct wlr_pointer_constraint_v1 *constraint);
@@ -196,6 +243,9 @@ static void destroypointerconstraint(struct wl_listener *listener, void *data);
 static void destroysessionlock(struct wl_listener *listener, void *data);
 static void destroykeyboardgroup(struct wl_listener *listener, void *data);
 static void destroytrackedpointer(struct wl_listener *listener, void *data);
+static void destroytrackedtablet(struct wl_listener *listener, void *data);
+static void destroytrackedtabletpad(struct wl_listener *listener, void *data);
+static void destroytrackedtablettool(struct wl_listener *listener, void *data);
 Monitor *dirtomon(enum wlr_direction dir);
 static void apply_input_settings_to_device(struct libinput_device *device);
 void focusclient(Client *c, int lift);
@@ -219,6 +269,15 @@ static void gestureholdend(struct wl_listener *listener, void *data);
 static void gpureset(struct wl_listener *listener, void *data);
 static void handlesig(int signo);
 static void inputdevice(struct wl_listener *listener, void *data);
+static void tabletnotifyaxis(struct wl_listener *listener, void *data);
+static void tabletnotifyproximity(struct wl_listener *listener, void *data);
+static void tabletnotifytip(struct wl_listener *listener, void *data);
+static void tabletnotifybutton(struct wl_listener *listener, void *data);
+static void tabletpadnotifybutton(struct wl_listener *listener, void *data);
+static void tabletpadnotifyring(struct wl_listener *listener, void *data);
+static void tabletpadnotifystrip(struct wl_listener *listener, void *data);
+static void tabletpadnotifyattach(struct wl_listener *listener, void *data);
+static void tabletpadnotifysurfacedestroy(struct wl_listener *listener, void *data);
 static int keybinding(uint32_t mods, uint32_t keycode, xkb_keysym_t sym, xkb_keysym_t base_sym);
 static void keypress(struct wl_listener *listener, void *data);
 static void keypressmod(struct wl_listener *listener, void *data);
@@ -341,6 +400,8 @@ struct wlr_pointer_constraints_v1 *pointer_constraints;
 static struct wlr_relative_pointer_manager_v1 *relative_pointer_mgr;
 static struct wlr_pointer_constraint_v1 *active_constraint;
 
+static struct wlr_tablet_manager_v2 *tablet_v2_mgr;
+
 struct wlr_cursor *cursor;
 /* Non-static so mousegrabber.c can access it */
 struct wlr_xcursor_manager *cursor_mgr;
@@ -361,6 +422,9 @@ struct wlr_output_layout *output_layout;
 static struct wlr_box sgeom;
 struct wl_list mons;
 static struct wl_list tracked_pointers; /* For runtime libinput config */
+static struct wl_list tracked_tablets;
+static struct wl_list tracked_tablet_pads;
+static struct wl_list tracked_tablet_tools;
 Monitor *selmon;
 static int in_updatemons;
 static int updatemons_pending;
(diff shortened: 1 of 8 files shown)


somewm.c
```diff
@@ -2227,6 +2443,583 @@ apply_input_settings_to_all_devices(void)
...
+	tool_v2 = get_or_create_tablet_v2_tool(tt->tablet, event->tool);
+	if (tool_v2) {
+		tablet_notify_proximity_in_for_cursor(tt->tablet, event->tool);
+		if (state == WL_POINTER_BUTTON_STATE_PRESSED)
+			wlr_tablet_v2_tablet_tool_notify_down(tool_v2);
+		else
+			wlr_tablet_v2_tablet_tool_notify_up(tool_v2);
+	}
+
+	wlr_seat_pointer_notify_button(seat, event->time_msec, BTN_LEFT, state);
```
JimmyCozza: This goes straight to the seat rather than through `buttonpress()`, so the client will see the click but somewm won't.

JimmyCozza:
This is one that's going to be hard for me to test since I don't have a tablet.  Did this end up having any overlap with PR #612?

Or are they completely separate?

shuber2:
Yes, that is a little bit uncomfortable, indeed. Maybe Wacom should sponsor you with a tablet :-)

Concerning #612 I discussed this there, too. To me it feels that this PR also adds infrastructure on top of which touch support can be implemented. See also the [last comment](https://github.com/trip-zip/somewm/pull/612#issuecomment-5216823227) by @zharsing.

 I may actually give touch support a try, too. Have some Thinkpad Yoga with Debian nearby.

shuber2:
@JimmyCozza I did an attempt to streamline the touch discussion in #707.

shuber2:
Well, I now added a commit to this branch that also adds touch support as described in #707. I have tested it on a Lenovo Yoga, which has both, a touch display and a tablet pen. Tested in krita and firefox, e.g., drawing, pinch to zoom, and scrolling.

It intends to replace or integrate #612.

shuber2:
And mouse tap-to-click added. I think this is a reasonable feature set for now. Let's leave it there.

Next steps:
* More testing
* Review the code
* Peer reviewing would be advisable

shuber2:
Fixed a few smaller bugs and inconsistencies in behavior.

JimmyCozza:
@shuber2 I don't want to stall this too long.

I'm tempted to merge this and watch for bug reports after the fact just because I really do not have a way to test this out myself.

Have you been running this yourself lately?  Have you noticed any bugs or concerns with it? 

What I can say definitively, is that it is at least inert when I run this branch as my daily driver, so my vote is merge and monitor.

shuber2:
I found that `map_from_region` for touch does not make sense, and sway also does not implement this. The manpage on sway-input was a bit misleading.

I am also about to cleanup a few comments.

Furthermore, I think about having an explicit test plan on a Lenovo Yoga Thinkpad with touch and a wacom tablet and an external monitor attached. If this goes through well, I think we are ready to merge as you said.

JimmyCozza:
I appreciate you working on this so long.  Sorry I can't be more help without a touch device right now

shuber2:
A few things got fixed:
- `map_from_region` now is only a tablet feature, like for sway.
- A couple of comments were too verbose and got trimmed.
- There was a bug in parsing "XxY" strings for `map_from_region` and `map_to_region` when "0x0" was passed. It was seen as one hex literal, rather two coordinates separated by "x". Adopted the sway parsing strategy.

shuber2:
# Manual test plan

Covers the tablet-v2 (#663) and wl_touch (#707) work in this PR. Target
hardware: a Lenovo Yoga (built-in touchscreen + integrated pen digitizer)
plus an external Wacom Intuos S (pressure only — no tilt, no pad ring/strip,
but it does have pad express keys), with an external monitor for the
multi-output scenarios.

Confirmed device names on this machine (`sudo libinput list-devices`), used
throughout below instead of a placeholder:

| Device | `Device:` name | `Id:` | Notes |
|---|---|---|---|
| Yoga integrated pen | `Wacom HID 5276 Pen` | `i2c:056a:5276` | non-USB (I2C) — expect `vid=0x0000 pid=0x0000` in somewm's `[tablet]` log line, that's expected, not a bug |
| Yoga touchscreen | `Wacom HID 5276 Finger` | `i2c:056a:5276` | pairs with the pen above |
| External Intuos S pen | `Wacom Intuos S Pen` | `usb:056a:0374` | real USB — expect real `vid`/`pid` in the log |
| External Intuos S pad | `Wacom Intuos S Pad` | `usb:056a:0374` | express keys, no ring/strip |

If your device names differ, substitute your own throughout.



## 1. Lenovo Yoga — built-in touchscreen

No config needed for 1.1-1.7 (C defaults).

- [x] **1.1** Single tap on a touch-aware app (Firefox/Krita canvas) — real touch delivered, no synthetic click on top.
- [x] **1.2** Single tap on a non-touch-aware app  — tap-to-click emulation fires: it receives a click and, if unfocused, gets focused/raised.
- [x] **1.3** Touch-scroll a long Firefox page.
- [x] **1.4** Two-finger pinch-to-zoom (Firefox page or Krita canvas).
- [x] **1.5** Two-finger gesture *on the same window again* (e.g. pinch) does not steal focus away from that window mid-gesture.
- [x] **1.6** Touch down on window A, then — while A's touch point is still active — tap window B: focus must **not** jump to B (`touch_point_active_elsewhere()` guard).
- [x] **1.7** Touch-drag to scroll in Firefox, lift, then immediately do a fresh short scroll gesture. Confirm the second gesture starts clean (no jump/glitch) and the page doesn't keep drifting on its own after you've lifted. (Don't use Krita's canvas for this one.)
- [x] **1.8** With only the internal panel enabled and no rule set, touch lands correctly on-screen (single-enabled-output default). Then add, reload, and confirm it still works pinned explicitly:
  ```lua
  awful.input.rules = {
      { rule = { type = "touch" },
        properties = { map_to_output = "eDP-1" } }, -- use your logged touch panel's output name
  }
  ```
  Remove this rule again before section 4.
- [x] **1.9** Disable emulation and confirm the non-touch-aware app from 1.2 no longer reacts to a tap:
  ```lua
  awful.input.rules = {
      { rule = { type = "touch" },
        properties = { emulate_pointer = false } },
  }
  ```
  Revert to `emulate_pointer = true` (or remove the rule) afterward.

## 2. Lenovo Yoga — integrated pen

- [x] **2.1** Bring the pen into proximity (hover) over the screen — pointer moves, no click.
- [x] **2.2** Tip down/up on a non-tablet-aware app — behaves like a left click (focus/raise if unfocused).
- [x] **2.3** Tip down/up in Krita — registers as a real tablet click, not a duplicated `wl_pointer` click.
- [x] **2.4** Draw a stroke in Krita/GIMP with varying pressure — line width/opacity responds to pressure.
- [x] **2.5** No physical eraser nib on the Yoga's pen, but its barrel button toggles eraser mode — confirmed working in Krita.
- [x] **2.6** Write a line of text in Xournal++.
- [x] **2.7** Annotate a PDF in Okular.
- [x] **2.8** Relative mode instead of absolute (named explicitly so this
      only affects the Yoga's pen once the Intuos is also connected later):
  ```lua
  awful.input.rules = {
      { rule = { type = "tablet-tool-pen", name = "Wacom HID 5276 Pen" },
        properties = { tool_mode = "relative" } },
  }
  ```
  Confirm the pen now moves the cursor relatively (like a mouse) instead of
  jumping to the absolute screen position matching where the pen touches
  the tablet surface. Remove the rule afterward (mouse/lens-type tools are
  always relative regardless of this setting — not applicable to a pen tip).

## 3. External Wacom Intuos S

No tilt, pad ring, or pad strip on this model — those code paths
(`tabletpadnotifyring`/`tabletpadnotifystrip`, tilt axis handling) stay
untested by this plan; note that explicitly rather than marking pass/fail.

Plug it in alongside the Yoga (both digitizers active at once). From here
on, any rule meant for only one pen needs an explicit `name` — `type =
"tablet-tool-pen"` alone now matches both the Yoga's and the Intuos's pen.

- [x] **3.1** Log shows `[tablet] new device name=Wacom Intuos S Pen size_mm=... vid=0x056a pid=0x0374` and `[tablet-pad] new device name=Wacom Intuos S Pad buttons=...` on plug-in.
- [x] **3.2** Draw in Krita with the Wacom pen — pressure works.
- [x] **3.3** Pad express-key buttons register in a tablet-aware app while the cursor is over its window.
- [x] **3.4** Pad focus follows the cursor: move the cursor to a different window/output, confirm pad enter/leave fires for the new window.
- [x] **3.5** Use the Yoga's built-in pen and the external Wacom pen in the same session without interference — draw a stroke with each, alternating.
- [x] **3.6** Pin the Wacom tablet to one specific output:
  ```lua
  awful.input.rules = {
      { rule = { type = "tablet", name = "Wacom Intuos S Pen" },
        properties = { map_to_output = "DP-8" }, -- your external monitor's output name
      },
  }
  ```
  Draw with the pen and confirm the cursor stays confined to that monitor
  even with the internal panel also enabled.
- [x] **3.7** Crop the tablet's active area to its top-left quadrant and confirm it's stretched to fill the whole mapped target (draw near the tablet's physical center - the cursor should already be near the target's bottom-right edge):
  ```lua
  awful.input.rules = {
      { rule = { type = "tablet", name = "Wacom Intuos S Pen" },
        properties = { map_from_region = "0x0 0.5x0.5" },
      },
  }
  ```
  Remove both rules (3.6, 3.7) afterward before section 4.

## 4. External monitor / hotplug

Make sure no `map_to_output`/`map_from_region` rules are left over from
sections 1 or 3 before starting this section.

- [x] **4.1** With only the internal panel enabled, touch and both pens map correctly (single-enabled-output default applies, no rules needed).
- [x] **4.2** Plug in the external monitor with **no** rules set. Per the documented caveat, the single-output default no longer applies once a second output is enabled - touch/tablet now span the *combined* layout. Confirm this matches that expectation (coordinates spread across both screens) rather than crashing or landing somewhere nonsensical.
- [x] **4.3** Add an explicit rule pinning the touchscreen back to the internal panel while the external monitor stays attached:
  ```lua
  awful.input.rules = {
      { rule = { type = "touch" },
        properties = { map_to_output = "eDP-1" } },
  }
  ```
  Confirm touch now stays pinned to the internal panel regardless of the
  external monitor.
- [x] **4.4** Unplug the external monitor again (rule from 4.3 still active) — confirm touch keeps working correctly on the internal panel.
- [x] **4.5** Remove the rule from 4.3, unplug/replug the external monitor into a different port if possible (so its connector name changes, e.g. `DP-8` → `DP-9`) — confirm a `map_to_output` rule using its make/model/serial id instead of the connector name still resolves it:
  ```lua
  awful.input.rules = {
      { rule = { type = "tablet", name = "Wacom Intuos S Pen" },
        properties = { map_to_output = "AOC U27E3U ZX3QBHA000360" }, -- your external monitor's make/model/serial id
      },
  }
  ```
- [x] **4.6** Realistic combined config — both pens pinned simultaneously, each matched the way that best fits its own stability (connector name for the internal panel, which never renumbers; make/model/serial for the external monitor, which can move ports on a dock):
  ```lua
  awful.input.rules = {
      { rule = { type = "tablet-tool-pen", name = "Wacom HID 5276 Pen" },
        properties = { map_to_output = "eDP-1" } },
      { rule = { type = "tablet", name = "Wacom Intuos S Pen" },
        properties = { map_to_output = "AOC U27E3U ZX3QBHA000360" } },
  }
  ```
  Confirm each pen draws only on its own intended screen.

## 5. Regressions

- [x] **5.1** Normal mouse/trackpad + keyboard use is unaffected.
- [x] **5.2** Existing click-to-focus/raise behavior for mouse clicks still works.
- [x] **5.3** A full session with touchscreen + both pens + external monitor connected simultaneously runs without crashes, hangs, or runaway log spam.

## Notes / follow-ups found during testing

(freeform — anything surprising, any log warnings seen, anything to file as
a separate issue rather than block this PR on; also the place to note that
tilt/ring/strip remain untested on this hardware)

shuber2:
@JimmyCozza I think I feel good about merging this now.

There is one remaining edge case open, I will investigate at some point: If a Wacom tablet is attached to a notebook with a tablet pen, then the Wacom tablets express buttons appear not to work, however, they do work if the Wacom tablet is the only device. This is probably not the case for many users. I would defer this for now.

zharsing:
@shuber2 

Hi Stefan,

I have tested on Lenovo P16 (touch and pen) and a secondary display connected over vkms+sunshine+artemis(android) to a Samsung S7+ (touch and pen)

`map_to_output` and `emulate_pointer` works as described. Touch scrolling and pich zoom are especially well behaving on capable apps.

I have ran into a couple of minor glitches:

Setup: Xournal++
Method:
- touch down the pen to wake the proxy (may vary per hardware)
- while in proximity and press and release the eraser button
- while still in proximity, press it again, observe the cursor
Observation:
- the cursor changes to eraser on the first release, yet functionally its in pen mode
- repeatedly pressing the eraser button toggles the cursor exactly the opposite of the expected behaviour
- moving out of proximity resets the state

Setup: open 2 clients: one with native touch and the other without (ie Sublime Text and Kitty) lay them out floating side by side
Method:
- hold modkey and touch the touch aware app
- cursor should change to indicate moving the client
- keep holding the modkey and touch the non-touch aware client
- watch the relative poistion of the cursor on the screen versus the touch point on the non-native client
- the vector between the 2 points is then applied to moving the touch aware client and the focus jumps to the other client
- keep playing with this and eventually you'll get a "mousegrabber already running" error from `awful/mouse/resize.lua`

Setup: having a non-touch aware app maximized (or otherwise overlapping) and a touch aware app floting on top of it (ie Sublime Text floating over Kitty)
Method:
- touch down on the floating app (ie. scroll) and hold while touching the non touch aware app in the beckground
Observation:
- the background app comes to front while the previously held touch point still gets delivered to the app in the background (observed via maximized transparent Kitty terminal and floating Sublime Text)


The touch to mouse emulation is not following the drag. (select text, resize window.. etc) the pencil works fine. It would be very interesting to have hotkeys for activating certain behaviour. Consider the following
Setup: Kitty terminal, floating
Idea:
- no key (or ctrl or shift) -> touch emulates mouse click and/or drag (ie. select text)
- hold ctrl or shift -> touch interpreted as scrolling
- hold modkey -> touch moves the window
- hold modkey and ctrl or shift -> touch resizes the window

The window management part could also be applied to pen workflow, which is already capable of performing some of the above. Might be as simple as adding a modifier to swap the left to the right mouse button?

Overall, I think its excellent - job well done!

- Zsolt


PS: In other systems its quite common to have long touch and hold result in context menu. This feature is actually available in compatible applications, yet the emulation might also address it. When I was looking for ideas and examples I came across a very interesting markup for mapping touch gestures: https://github.com/swaywm/sway/issues/1904

Might be wrong, but it could be as simple as adding a signal for touch frame event..

shuber2:
@zharsing Thanks for the extensive feedback. Will have a look at it. The mouse emulation for touch is very basic, and doing more is extensive work.

Do you think the current branch can be merged as is, or do you think there are blockers that need to be resolved first?

zharsing:
@shuber2 indeed, the proper mouse emulation is rather tricky, which is precisely why I did not go into that rabbit hole (yet).

I don't see any serious issue or any regression, while on the other hand this change adds precious features. There is certainly room for improvement, yet I think this is a good addition already, as is.

Disclaimer: I am using the main/2.0 branch, only built this to test and I won't be running it on the daily. My superficial review is strictly from user perspective, have not looked into the code - its quite an overwhelming amount you added.. :) Let alone the sheer size of somewm.c..

Thumbs up from me. 

--Zsolt

shuber2:
Great. And maybe this can be a baseline implementation for release/1.4 and more feature-heavy dev can be in the 2.0 branch.

JimmyCozza (approved):
Thank you both @shuber2 and @zharsing 

Create followup issues if you notice anything else out of place here.

PR #704, opened 2026-08-19
fix(popup): fix position and z-order of popups on release/1.4

shuber2:
## Description

It fixes a bug where popup menus and titlebar menus were offset. See here:

menu-bad

The offset was actually the client's border width and title bar height. The underlying reason was that menus were placed on the clients (outer) `scene` rather than its (inner) `scence_surface`. This is how it looks correctly with this fix:

menu-good

## Test Plan

Manual test, see screenshots.

## Checklist

### DIFF
--- a/objects/client.h
+++ b/objects/client.h
@@ -158,6 +158,10 @@ struct client_t
     struct wlr_scene_tree *scene;
     /** Scene surface node */
     struct wlr_scene_tree *scene_surface;
+    /** Popup parent tree: tracks scene_surface's position but is exempt
+     * from client_get_clip()'s content clip (mirrors LayerSurface::popups),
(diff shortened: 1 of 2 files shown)


shuber2:
Oh, firefox now displays a bug where its menu's are cropped on the left-top side. Have to investigate.

shuber2:
Fixed. Claude generalized the old version (using scene_surface instead of scene) by introducing a dedicated popup surface that is positioned like scene_surface.

shuber2:
And fixed another bug: When popups extend beyond a clients surface, the client border was drawn above the popup. Raise z-order of popups to top.

JimmyCozza (approved):
Thanks @shuber2!

PR #705, opened 2026-08-19
fix(popup): fix position and z-order of popups on main

shuber2:
## Description

This is the port of #704 to the main branch

## Checklist


shuber2:
Now also fixes the firefox issue in analogy to #704.

shuber2:
And fixed another bug: When popups extend beyond a clients surface, the client border was drawn above the popup. Raise z-order of popups to top.

JimmyCozza (approved):
Thank you!

PR #714, opened 2026-08-25
fix(build): libpam fallback detection for Debian

shuber2:
## Description

The lockscreen did not work on Debian machines. The reason was that meson would not find pam, although the pam dev package was installed. However, upstream PAM does not ship pkg-config files, and so it depends on the distro to ship one. Debian does not. Hence, we need a fallback mechanism here.

## Test Plan

Manual test on Debian. Confirmed on Debian Trixie on two different machines.

## AI Usage

Debugged and implemented with the help of Claude.

## Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -204,8 +204,26 @@ message('Using wlroots version: ' + wlroots_version)
 
 # PAM authentication (optional, for lock screen)
 pam_option = get_option('pam')
-pam_dep = dependency('pam', required: pam_option)
-have_pam = pam_dep.found()
+# There are three cases: If pam_option is disabled, we do not look for and do
+# not build against PAM. If pam_option is enabled, we require PAM to the found.
+# If pam_option is auto, we opportunistically enable it if found.
+have_pam = false
+if not pam_option.disabled()
+  pam_dep = dependency('pam', required: false)
+  if not pam_dep.found()
+    # Some distributions (e.g., Debian) do not ship PAM with pkg-config/cmake
+    # files. Do a fallback scan.
+    pam_cc = meson.get_compiler('c')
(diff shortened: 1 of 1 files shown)


JimmyCozza (commented):
@shuber2 Thanks for finding this.

I do have 1 question about this for you.  Now that dependency() is called with `required:false`, does -Dpam=disabled still turn PAM off where pam.pc does exist?  I would just like to make sure that we keep that path working for anyone who locks with swaylock and don't want libpam linked into the compositor.

shuber2:
I was just about to write why the logic is correct, until I saw it was not. Oh man, three-valued logic fail.

Anyway, I took the opportunity to add some comment on the intended logic and reorganized the case distinction. Now it is evident that `pam_option.disabled()` leads to `have_pam = false`. The `pam_dep` logic is entirely opportunistic. The case for `pam_option.enabled()` is a single post-hoc check.

Sorry I did not realize right ahead.

JimmyCozza (approved):
Looks great!

Thanks @shuber2

shuber2:
By the way, this discussion motivated me to add a pam use-flag to the somewm ebuilds in the [somewm-gentoo-overlay](https://github.com/shuber2/somewm-gentoo-overlay), so users can choose.

shuber2:
@JimmyCozza Ah, it might have been good to cherry-pick this one for the 2.0.0 release.

JimmyCozza:
On it.

2.0.1 coming in quickly.

On the topic of 2.0, do you have a strategy for how to handle packaging on gentoo?  I'm trying to make sure I have the best path for AUR.  It's a bit of a pain, I know.

PR #717, opened 2026-08-27
fix(systemd): use command -v instead of hash for POSIX compliance

mcepl:
hash is a bashism and not guaranteed in POSIX sh (dash etc.). command -v is the POSIX-standard way to check for command existence, fixes ShellCheck SC2143.

## AI Usage

It could be a good philosophical discussion: where is the border between “normal” tools (e.g., I used `checkbashism`) and what we call now AI tools. What’s the difference?

## Checklist

### DIFF
--- a/contrib/systemd/somewm-session.sh
+++ b/contrib/systemd/somewm-session.sh
@@ -26,7 +26,7 @@ if [ -n "${MANAGERPID:-}" ] && [ "${SYSTEMD_EXEC_PID:-}" = "$$" ]; then
 fi
 
 # Make sure systemd is available
-if ! hash systemctl >/dev/null 2>&1; then
+if ! command -v systemctl >/dev/null 2>&1; then
     echo "systemd not found. Run somewm directly instead." >&2
     exit 1
 fi
@@ -59,7 +59,7 @@ systemctl --user set-environment \
     XDG_SESSION_TYPE=wayland
 
 # D-Bus activation environment is independent from systemd
-if hash dbus-update-activation-environment 2>/dev/null; then
+if command -v dbus-update-activation-environment 2>/dev/null; then
     dbus-update-activation-environment --all
 fi
 


contrib/systemd/somewm-session.sh
```diff
@@ -59,7 +59,7 @@ systemctl --user set-environment \
     XDG_SESSION_TYPE=wayland
 
 # D-Bus activation environment is independent from systemd
-if hash dbus-update-activation-environment 2>/dev/null; then
+if command -v dbus-update-activation-environment 2>/dev/null; then
```
JimmyCozza, line 62: Correct me if I'm wrong, but this is actually going to print the resolved path to stdout, whereas hash printed nothing.

So I think it'll need to have the stdout redirect too:

`if command -v dbus-update-activation-environment >/dev/null 2>&1; then`

JimmyCozza (changes requested):
I think it's a good change, I'd be happy with `command -v`

Though I am kind of curious about the very specific wording in your commit message.
I'm not saying I'm perfect here in my understanding by any stretch, but 

> hash is a bashism and not guaranteed in POSIX sh (dash etc.).

I guess I'm confused on such a hardline phrasing.  I was under the impression that hash WAS guaranteed in a POSIX shell: https://pubs.opengroup.org/onlinepubs/9799919799/utilities/hash.html

Not that it matters a ton, but dash has it with exactly the POSIX synopsis (https://manpages.debian.org/testing/dash/dash.1.en.html).

> fixes ShellCheck SC2143
https://www.shellcheck.net/wiki/SC2143

What does `Use grep -q instead of comparing output with [ -n .. ]` have to do with this PR?

Can you clean up the commit message a bit?

JimmyCozza:
As for the AI clarification.  

I'm not super interested in a philosophical debate about where the line between AI or not AI is.  I deal with that banality enough at work.

I don't personally see a difference between someone submitting a PR where they

a) generated a change via LLM and do not understand it, or cannot explain the reasoning/solution, or
b) used a standard CLI tool to generate a change and do not understand it, or cannot explain the reasoning/solution.

Either way the whole burden of validation lands on the maintainers. That's what the policy is actually aimed at 
>It's the people, not the tools.

The addition of the AI_POLICY.md is that "Yes, we do accept LLM assisted code.  Provided you follow this guideline"

mcepl:
OK, you are right, I (and `checkbashism`) am wrong. `hash` is POSIX.

Issue #752, opened 2026-09-02
Forward-port tablet-and-touch from release/1.4


shuber2:
### What feature would you like?

With #692 being merged into release/1.4, we want to have the same for the main branch for the 2.x releases. This issue is here to track this step.

### Forward porting stategy

The main purpose here is to have the 6 commits from release/1.4 being forward ported, with no new features or fixes, but forward ports as close as possible.  I have a first proof of concept branch [tablet-touch-fwdport-2.x](https://github.com/shuber2/somewm/tree/feat/tablet-touch-fwdport-2.x) ready.

Next steps:
- Check the code. Forward port was done by Claude. This worked quite well, with only little intervention on the commit messages. But it has not been tested.
- Repeat the test protocol as in #692.
- Use it for a while in daily setup.

### Post forward-port activities

Once the transition to main branch was successful, I would investigate the points raised by @sicelo  in #524 and by @zharsing in #692. Some point to non-trivial features, like advanced mouse emulation. Some smell like bugs or issues within the current feature set. The former may or may not be backported to release/1.4, the latter should.

@JimmyCozza Any thoughts on this? Sounds like a plan?

JimmyCozza:
Sounds good to me.

I did a single PR where I cherry picked a handful of commits in to main last night.

I only hesitated on the touch in case you wanted to wait for any reason.

If you feel good about it let's get it to main.

shuber2:
@JimmyCozza I am a bit confused. The commits already landed in main? But these are not the commits from my [tablet-touch-fwdport-2.x](https://github.com/shuber2/somewm/tree/feat/tablet-touch-fwdport-2.x) branch, right?

JimmyCozza:
Yeah they're on main now via #758.

I just cherry-picked the commits that I could and cleaned up the ones that needed to be adapted to the main branch.  
I'll be honest, I had just been cherry-picking commits from release/1.4 -> main nightly and caught those ones in my flow since it was easier to do them all rather than leave the touch ones out.

Do they look OK on main?

shuber2:
Yeah, they look good. But then I remove my forward port branch.

PR #753, opened 2026-09-02
fix(input): allow pointer emulation even outside clients

sicelo:
## Description
Allow elements such as wibar and title bar to also receive emulated pointer events

## Test Plan
Tested on my device mentioned in https://github.com/trip-zip/somewm/issues/524

## AI Usage
None

## Checklist

### DIFF
--- a/somewm.c
+++ b/somewm.c
@@ -2767,14 +2767,11 @@ touchnotifydown(struct wl_listener *listener, void *data)
 	input_map_coords(&s, event->x, event->y, &lx, &ly);
 	xytonode(lx, ly, &surface, &c, NULL, NULL, NULL, &sx, &sy);
 
-	if (!surface)
-		return;
-
 	/* wlroots itself refuses to create a touch point (logged as an ERROR)
 	 * for a client with no bound wl_touch resource, so only forward the
 	 * raw event to clients that can actually use it. Everyone else gets
 	 * an emulated click instead, if enabled. */
-	if (!touch_client_needs_pointer_emulation(surface)) {
+	if (surface && !touch_client_needs_pointer_emulation(surface)) {
 		/* Touch-aware clients get real, raw touch - no synthetic
 		 * wl_pointer click (see touch_emulate_click()'s comment for
 		 * why). They still deserve focus/raise on a deliberate tap


somewm.c
```diff
@@ -2767,8 +2767,11 @@ touchnotifydown(struct wl_listener *listener, void *data)
 	input_map_coords(&s, event->x, event->y, &lx, &ly);
 	xytonode(lx, ly, &surface, &c, NULL, NULL, NULL, &sx, &sy);
 
-	if (!surface)
+	if (!surface) {
```
JimmyCozza: This adds a second touch_emulate_click() call, but the else if (s.emulate_pointer) at the bottom of this function is already that call. 

If you delete this if (!surface) block entirely and change the condition below to if (surface && !touch_client_needs_pointer_emulation(surface)), the no-surface case falls through to the existing else if on its own, and the comment above it ("everyone else gets an emulated click") stays accurate.

somewm.c
JimmyCozza: add the surface && check here IMO

shuber2:
That looks nice, however, I did not had time to test it yet.

I would suggest to follow the minimal patch as suggested by @JimmyCozza. Like this:

```diff
diff --git a/somewm.c b/somewm.c
index 3e0ac17..665778f 100644
--- a/somewm.c
+++ b/somewm.c
@@ -2767,14 +2767,12 @@ touchnotifydown(struct wl_listener *listener, void *data)
        input_map_coords(&s, event->x, event->y, &lx, &ly);
        xytonode(lx, ly, &surface, &c, NULL, NULL, NULL, &sx, &sy);

-       if (!surface)
-               return;
-
        /* wlroots itself refuses to create a touch point (logged as an ERROR)
         * for a client with no bound wl_touch resource, so only forward the
         * raw event to clients that can actually use it. Everyone else gets
-        * an emulated click instead, if enabled. */
-       if (!touch_client_needs_pointer_emulation(surface)) {
+        * an emulated click instead, if enabled - including taps that land
+        * on no surface at all, e.g. the wibar or a titlebar. */
+       if (surface && !touch_client_needs_pointer_emulation(surface)) {
                /* Touch-aware clients get real, raw touch - no synthetic
                 * wl_pointer click (see touch_emulate_click()'s comment for
                 * why). They still deserve focus/raise on a deliberate tap
```

Was thinking about exposing limitations of mouse emulation for touch to non-surface areas through this change. For instance, we do not have enter and leave events. But I think, it is fine. 

By the way, the tablet code in `tabletnotify()` always sends `buttonpress()`. (However, here prior pen proximity events already give cursor motions leading to enter and leave events.)

If this gets merged then I would take it along with #752, too.

sicelo:
Thanks for review. Changes implemented.

PR #755, opened 2026-09-03
fix: restore client tags and selected tags across a hot-reload

### LINKED ISSUE #731: Client tag membership is not preserved across a hot-reload
`awesome.restart()` drops which tags a client was on. Everything re-tags from scratch and lands on the selected tag, so a user's session collapses onto tag 1. Unlike #730 this hits single-monitor setups too - it is missing state, not a wrong flag.

`tag_snapshot_t` (`luaa.c:5161`) is a single field, `char *name`. Tag membership lives only in `tag_t.clients` (`objects/tag.h:63`) - there is no per-client tag list - and `luaa.c:5337` zeroes it. The comment at `:5333` says so outright: "Tags are recreated from rc.lua defaults, not restored from snapshots." `client_snapshot_t` doesn't carry tags either.

After rc.lua reloads, `luaa.c:5601` emits `request::manage` with context `"restart"`, but `ruled.client.apply` takes only `c` and ignores the context (`lua/ruled/client.lua:185`), so rules run as if the client were new. Anything unmatched falls through to `awful.permissions.tag` and `c:to_selected_tags()` (`lua/awful/permissions/init.lua:337`), which resolves to the freshly selected tag 1.

Fix shape: snapshot per-client tag names and per-screen selected tag names, then re-apply by name after rc.lua has created the tags, instead of relying on rules.

Open question: what to key the restore on. Matching by name is the obvious choice but mis-restores silently when a config has duplicate tag names or renames tags between reloads. Matching by index-within-screen survives renames but breaks when the config changes its tag count. Neither is right for a config that does both, and the fallback when a saved tag has no match after reload needs deciding too.

`tests/restart/tags-after-restart.sh:38` claims to assert this, but `tests/rc.lua:25` creates exactly one tag, so both sides trivially resolve to the same tag and it passes with the bug present.

Line numbers are `release/1.4`; same on main at different offsets.

JimmyCozza:
## Description
Closes #731. 

A hot-reload rebuilt every tag from rc.lua and dropped which tags each client was on and which tags were viewed. 

The compositor now records both before the old state closes, matches each tag by name per screen after rc.lua runs, emits request::tag with every matched tag before the rules, and restores the selection before the rules so a client left with no match lands on it. Renamed and runtime-created tags fall through to the rules. Modifies lua/awful/permissions/init.lua to apply hints.tags; documented in DEVIATIONS.md.

### DIFF
--- a/DEVIATIONS.md
+++ b/DEVIATIONS.md
@@ -187,7 +187,7 @@ These modifications to AwesomeWM's Lua libraries were necessary for Wayland comp
 | `awful/widget/tasklist.lua` | Icon slot falls back to `awful.client.get_icon_path(c)`; templates can receive a file path string where upstream always passes a surface | Same missing Wayland icon protocol |
 | `awful/widget/keyboardlayout.lua` | `next_layout`/`set_layout` wrap-around off-by-one fixed | Upstream cycles to an out-of-range group index |
 | `awful/widget/layoutlist.lua` | `source.for_screen` returns `{}` for a nil screen instead of asserting | Screens can be absent mid hot-reload |
-| `awful/permissions/init.lua` | Layer surface keyboard focus handlers; `request::tag` guards a nil `transient_for.screen`; hosts the focus restore and tag persistence handlers (see SomeWM-Only Features) | Wayland layer-shell, output hotplug and hot-reload have no X11 equivalent |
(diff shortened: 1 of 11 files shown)


JimmyCozza:
* Not quite ready.  Found 2 more things I'd like to roll into this...but I'm tired.

When you restart now, clients keep their tags and screens, but

a) tags don't keep their layouts
b) in floating, clients seem to grow in height.  Unnoticeable in 1 restart, but unmistakable if you restart multiple times quickly, you can watch it grow relatively fast.

PR #756, opened 2026-09-03
feat(input): apply libinput rules on tablet and touch devices

sicelo:
Enable the use of global and per-device input rules on tablet, tablet pad, and touch devices.

## Description
Allows applying input rules to the new device classes (tablet, pad, and touch), e.g. to disable a device or set other properties

## Test Plan
Tested on-device, touch only. Tablet and pad NOT tested since I do not have access to such a device.

The following `$HOME/.config/somewm/rc.lua` snippet:
```
awesome.connect_signal("dpms::on", function(ev)
    awful.input.rules = { { rule = { name = "TSC2005 touchscreen" }, properties = { send_events_mode = "enabled" } }, }
end)

awesome.connect_signal("dpms::off", function(ev)
    awful.input.rules = { { rule = { name = "TSC2005 touchscreen" }, properties = { send_events_mode = "disabled" } }, }
end)
```

**Result**
The touchscreen is disabled while the display is off, so interacting with the touchscreen at that point does not wake the display

## AI Usage
None

## Checklist


sicelo:
@shuber2 your review/comment is also important for this one, since it's a follow-up to your tablet and touch patch series

shuber2:
Ah, interesting. Yes, you address a real gap here. I only have a few seconds and then go for a short vacation trip until Monday, here are a few comments:

  - The commit message should say something like current rules are only resolved during touch and tablet events, but `apply_input_settings_to_device()` was never called, and this function did not handle touch and tablet devices.
  - Maybe a comment that tablet pens are not relevant here.
  - The example above points to another thing, later relevant for documentation: Setting `awful.input.rules` is probably not the pattern we want in general, but we need a update-rule-by-name lua pattern.
  - I am just thinking about the guard logic here. Is it possible that `tracked_touches` and friends contain non-libinput devices in the `apply_input_settings_to_device()` loops? Because the pointer loop has no `wlr_libinput_get_device_handle()`.

Maybe you can look at this? I am back next week.

sicelo:
Thanks for the review.

> commit message should say something like current rules are only resolved during touch and tablet events, but apply_input_settings_to_device() was never called, and this function did not handle touch and tablet devices

As it stands (before this patch), I am quite sure all rules are not applied for the new class of devices. At least they do not take effect on my touchscreen, regardless of how I define them in the config file.

Having looked at the existing code further (after the review), it seems the assumption may have been made that `resolve_<type>_settings()` makes the settings take effect. It does not. It only evaluates what settings should be in effect when the function is called. Whether those are actually in effect is a different matter, as they only take effect once `apply_input_settings_to_device()` is called. This is why the rules never applied even though `resolve_<type_settings()` is called multiple times.

TLDR; I think the commit message is fine.

> Maybe a comment that tablet pens are not relevant here

Will add this. I don't have access to tablet and accessories so I am not sure what capabilities really exist and how they could be supported

> Relevant for documentation: Setting awful.input.rules is probably not the pattern we want in general, but we need a update-rule-by-name lua pattern

I think this would be better as its own standalone patch. The existing pattern does work and can be discovered intuitively based on the existing documentation, but I can agree that implementing named rules could shorten it.

> the guard logic here. Is it possible that tracked_touches and friends contain non-libinput devices in the apply_input_settings_to_device() loops? Because the pointer loop has no wlr_libinput_get_device_handle()

Not very likely afaict, and I believe `apply_input_settings_to_device()` can handle a `NULL` device without issues. However, I can and will absolutely add `wlr_input_device_is_libinput()` safeguards to handle edge cases

shuber2:
Sorry for the long delay, but I could not make it earlier.

First, I think that the commit itself is fine.

What follows are a few thoughts that may add to the context and might change the commit message.

**1. Whether rules are applied.**

I would argue that rules _are_ applied prior this commit. More precisely, all rules of the test protocol listed in #692 are effective. However, the mechanism by which they are effective is, though, a different one: `resolve_input_settings_for_types()` is evaluated fresh on every tablet/touch event (e.g. `touchnotifydown()`, `tabletnotifyproximity()`), not pushed once via `apply_input_settings_to_device()` as this commit does for libinput-level device settings.

Your PR is an important one for two reasons, in my opinion:

  1. Your example of setting `send_events_mode` is of a general type that is not covered by the above mechanism.

  2. What bothers me are the computational costs of the mechanism I explained above. I would rather offload the computational cost to build the mapping "matrix" only upon monitor or input device events and keep the computational path in `touchnotifydown()` et cetera low. And here also your mechanism will be the way to go.

Hence, I would suggest the following commit message:

```
feat(input): apply libinput device settings to tablet, pad, and touch

Rules affecting map_to_output/tool_mode only worked as they are
resolved fresh on every event via resolve_input_settings_for_types().

libinput-level settings (send_events_mode, tap_to_click, accel_speed,
etc.) did not: apply_input_settings_to_device() was only ever called
for pointer devices, at creation and on rule changes. Extend it to
tablet, tablet pad, and touch devices. Tablet pens are not separate
libinput devices and are unaffected.

Signed-off-by: Sicelo A. Mhlongo <absicsz@gmail.com>
```

**2. Necessity of the guard logic.**

Note that `device_from_touch()` of wlroots has `assert(wlr_touch->impl == &libinput_touch_impl)`, which would then lead to an abort. But even if asserts are stripped, `wl_container_of()` does not return `NULL` for `dev`: it just computes an offset from `wlr_touch`, so for a non-libinput device it hands back a tainted pointer instead. And `return dev->handle` in `wlr_libinput_get_device_handle()` then dereferences that pointer.

Any backend other than libinput (e.g. wlroots' `wayland` or `headless` backend, both commonly used for nested development/testing) can produce a tracked touch/tablet/pad device that would trip this on the very first rule (re)application. `tracked_touches` and friends are populated unconditionally at hotplug time regardless of backend, unlike `tracked_pointers`, which is only ever populated inside the same `wlr_input_device_is_libinput()` check.

I think the guard you added is really not cosmetic.

Maybe, in order to keep the code a bit tighter, one could factor out the repeated pattern of checking as follows:

```c
static void
apply_input_settings_if_libinput(struct wlr_input_device *base)
{
	struct libinput_device *device;
	if (wlr_input_device_is_libinput(base)
			&& (device = wlr_libinput_get_device_handle(base)))
		apply_input_settings_to_device(device);
}
```

which would turn each of the six call sites into a one-liner, e.g. `apply_input_settings_if_libinput(&tc->touch->base);`. And then `apply_input_settings_to_all_devices()` could look like this:

```c
void
apply_input_settings_to_all_devices(void)
{
	TrackedPointer *tp;
	TrackedTouch *tc;
	TrackedTablet *tt;
	TrackedTabletPad *ttp;

	wl_list_for_each(tp, &tracked_pointers, link)
		apply_input_settings_to_device(tp->libinput_dev);

	wl_list_for_each(tc, &tracked_touches, link)
		apply_input_settings_if_libinput(&tc->touch->base);

	wl_list_for_each(tt, &tracked_tablets, link)
		apply_input_settings_if_libinput(&tt->tablet->base);

	wl_list_for_each(ttp, &tracked_tablet_pads, link)
		apply_input_settings_if_libinput(&ttp->pad->base);
}
```

(`tracked_pointers` keeps its own loop since `TrackedPointer` only stores the already-resolved `libinput_dev`, not the `wlr_pointer` itself.)

And the pen comment, right in `createtablet()`:

```c
	apply_input_settings_if_libinput(&tablet->base);
	/* Tablet pens are not separate libinput devices; their settings live
	 * on the tablet above, so there is nothing more to apply here. */
```

sicelo:
It appears that we are completely misaligned (`resolve_input_settings_for_types` vs. `apply_input_settings_to_device`). However landing the fix in somewm is of higher importance for me, so I'll stop here and just write what you say so we can merge this without any further delay.

shuber2:
Thank you for being pragmatic.

But I really do not want to leave you with bad feelings and I would be happy if we can clarify disagreements. Unfortunately, I am not sure where these actually lie at.

So what I mean by "map_to_output rule does apply" is maybe best illustrated in the code below from `somewm.c`. The `resolve_touch_settings()` does what I mean with "map_to_output rule does apply".

```C
static void
touchnotifydown(struct wl_listener *listener, void *data)
{
      struct wlr_touch_down_event *event = data;
      struct InputSettings s;                              // local, stack-allocated
      ...
      resolve_touch_settings(event->touch, &s);            // freshly resolved, every single call
      input_map_coords(&s, event->x, event->y, &lx, &ly);  // map_to_output applied right here
      ...
```

JimmyCozza:
Thanks to both of you, this is a real gap and I want to get it merged. 

Reading the thread, I think you are describing two halves of the same thing, the mapping rules already took effect because every touch and tablet event resolves them fresh, and the libinput-level settings never did because apply_input_settings_to_device() was pointer-only. The commit message as it stands now says exactly that, so I consider that part done.

One thing I noticed while reading the code that I would like to fix before merging.  

get_input_device_type() returns "touchpad" or "pointer" for any libinput device

so a touchscreen or  tablet going through apply_input_settings_to_device() resolves rules as if it were a pointer. That  means a rule with type = "touch" or type = "tablet" and send_events_mode still will not apply, and an existing type = "pointer" rule will now also hit tablets and touchscreens. Would it work to let the caller pass the device class into the resolver, so createtouch() passes "touch", createtablet() passes "tablet", createtabletpad() passes "tablet-pad", and createpointer() keeps the current classifier?

Two smaller things from the earlier review that are not in the diff yet: the three create*() sites could use the new apply_input_settings_if_libinput() helper instead of repeating the guard, and the pen comment next to the tablet call. The guard itself is definitely needed, wlroots asserts in device_from_touch() for non-libinput devices, so I would rather not see it dropped.

@sicelo I understand if you're frustrated with this and don't really want to do anymore.  If so, LMK and I'll merge this and then get the updates pushed for you to review later?

sicelo:
I will work on that. Thanks for the review. Appreciated

sicelo:
Please have a look at the latest iteration - I reworked things a bit:

- instead of storing and passing around a reference to `libinput_device`, pass around `wlr_input_device` instead, since from this it is trivial to obtain `libinput_device` when needed, and **also** determine the device type (with the help of the reworked `get_input_device_type()`. Bonus point is that it means you can also apply libinput settings/properties to any `wlr_input_device` now
- with the above changes, `resolve_input_settings()` is no longer necessary, since we can just directly call `resolve_input_settings_for_types()` from within  `apply_input_settings_to_device()`, with the correct device name and type
- the suggested helper, `apply_input_settings_if_libinput()` is also not needed anymore since the `wlr_input_device_is_libinput()` guard can now be easily integrated within `apply_input_settings_to_device()` itself

Updated the earlier rule to:

```
awesome.connect_signal("dpms::on", function(ev)
    awful.input.rules = { { rule = { type = "touch" }, properties = { send_events_mode = "enabled" } }, }
end)

awesome.connect_signal("dpms::off", function(ev)
    awful.input.rules = { { rule = { type = "touch" }, properties = { send_events_mode = "disabled" } }, }
end)
```
and it still works.

Regarding the pens comment - I have added the comment to clearly specify that this affects `wlr_input_device`, which should make it clear that pens and all other tablet tools are out of scope of the MR (yes pens are not the only devices involved here).

JimmyCozza (approved):
Thanks @sicelo and @shuber2

Going to merge this and we can followup later.

Issue #765, opened 2026-09-08
supporting IME for Chinese input (impleting wayland's text-input-v3)


myf:
### What feature would you like?

Wayland handles input methods in an unified way, replacing the old X ecosystem with separate implementations for GTK and Qt etc. Input methods for non-latin based scripts such as `ibus` and `fcitx5` uses these new api. This API needs to be implemented by the window manager itself. I'd like to be able to type Chinese under somewm. 

sway is a prior art that had implemented `text-input-v3`. I have explored using sway's `/input/text_input.c` as a base and generated the appropriate code for somewm's text-input-v3 implementation. However the dillenma is that it is indeed llm generated that carries a different risk in terms of future maintenance. I will leave the PR here but not expecting it to be merged, but rather a conversation starter.

### Use case

the problem solved here is being able to input non-latin based scripts via `ibus` or `fcitx5`.

### Proposed API or rc.lua shape

```lua

```

### Prior art

sway has inplemented these two APIs in `/input/text_input.c` and I was able to use the Chines input properly. 

### Preflight checklist

- [x] I have searched existing issues and this feature is not already requested.
- [x] This is a new feature, not a bug fix or an AwesomeWM 4.4 parity gap.

PR #766, opened 2026-09-08
Adding text_input_v3 and text_input_v2 in order to add IME support

myf:
to Chinese input methods (fctix5, ibus).

`zwp_text_input_v3` for clients and `zwp_input_method_v2` are implementd.

ref: #765

Changes
    - input_method.c / input_method.h — the relay. Adapted from sway's sway/input/text_input.c (MIT), with three departures for somewm: a single global relay rather than per-seat; focus tracked via seat->keyboard_state.events.focus_change instead of patching the nine wlr_seat_keyboard_notify_enter call sites; and IME popups parented onto the existing client_t::popups / LayerSurface::popups trees, which avoids sway's scene_descriptor machinery.
    - input.c — routes keys and modifiers to the input method's keyboard grab in keypress() / keypressmod(), after keybinding handling so compositor bindings still win. Records the originating virtual keyboard on KeyboardGroup so an IME's own injected keys aren't fed back to it.
    - wlr_compat.h — 0.20 renamed the manager new_text_input / new_input_method signals; macros keep 0.19 building.
    - somewm.c, somewm_types.h, meson.build — manager creation, relay init/finish, struct field, source entry.

 AI disclosure: written with Claude Code. The relay module was AI-drafted against sway as a reference and reviewed/modified by me; a NULL-deref crash in the keyboard-grab destroy handler was diagnosed from coredumps and fixed.

## Checklist

### DIFF
--- a/input.c
+++ b/input.c
@@ -61,8 +61,10 @@
 #include "objects/signal.h"
 #include "event.h"
 #include "xkb.h"
+#include <wlr/types/wlr_input_method_v2.h>
 #include <wlr/types/wlr_virtual_keyboard_v1.h>
 #include <wlr/types/wlr_virtual_pointer_v1.h>
+#include "input_method.h"
 
 /* macros */
 #define CLEANMASK(mask) (mask & ~WLR_MODIFIER_CAPS & ~WLR_MODIFIER_MOD2)
@@ -1138,6 +1140,7 @@ keypress(struct wl_listener *listener, void *data)
 	int handled = 0;
 	uint32_t mods;
 	xkb_keysym_t base_sym;
+	struct wlr_input_method_keyboard_grab_v2 *kb_grab;
 	/* This event is raised when a key is pressed or released. */
 	KeyboardGroup *group = wl_container_of(listener, group, key);
 	struct wlr_keyboard_key_event *event = data;
@@ -1221,6 +1224,16 @@ keypress(struct wl_listener *listener, void *data)
 	if (some_is_lua_locked())
 		return;
 
+	/* An input method holding the keyboard grab consumes what the compositor
+	 * did not. Checked after keybindings so bindings still win over the IME. */
+	if ((kb_grab = input_method_get_keyboard_grab(group))) {
+		wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+				&group->wlr_group->keyboard);
+		wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
+				event->time_msec, event->keycode, event->state);
+		return;
+	}
+
 	wlr_seat_set_keyboard(seat, &group->wlr_group->keyboard);
 	/* Pass unhandled keycodes along to the client. */
 	wlr_seat_keyboard_notify_key(seat, event->time_msec,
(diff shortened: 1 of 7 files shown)


input_method.c
```diff
@@ -0,0 +1,684 @@
...
+input_popup_destroy_scene_tree(InputPopup *popup)
+{
+	if (popup->scene_tree) {
+		wlr_scene_node_destroy(&popup->scene_tree->node);
+		popup->scene_tree = NULL;
+	}
+}
+
+static void
+handle_focused_surface_unmap(struct wl_listener *listener, void *data)
```
JimmyCozza, line 174: The popup tree is a child of c->popups, which unmapnotify already destroys. Does this run after that and destroy it a second time?

input_method.c
```diff
@@ -0,0 +1,684 @@
...
+
+static void
+handle_im_keyboard_grab_destroy(struct wl_listener *listener, void *data)
+{
+	/* wlroots emits this signal with NULL data, despite wlr_input_method_v2.h
+	 * annotating it "struct wlr_input_method_keyboard_grab_v2". Take the grab
+	 * from the input method instead; the input method itself may already be
+	 * gone if the grab is being torn down as part of its destruction. */
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
+			im_relay.input_method ? im_relay.input_method->keyboard_grab : NULL;
```
JimmyCozza, line 376: On wlroots 0.19 the IM destroy fires before the grab destroy, so this is NULL right when fcitx5 exits. Should the grab be stored at grab time instead?

input_method.c
```diff
@@ -0,0 +1,684 @@
...
+	}
+}
+
+static void
+handle_im_grab_keyboard(struct wl_listener *listener, void *data)
+{
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab = data;
+
+	wlr_input_method_keyboard_grab_v2_set_keyboard(keyboard_grab,
+			wlr_seat_get_keyboard(seat));
```
JimmyCozza, line 395: When fcitx5's virtual keyboard was the last one used, isn't this the IME's own keyboard? Why not the physical group?

input_method.c
```diff
@@ -0,0 +1,684 @@
...
+	TextInput *text_input = wl_container_of(listener, text_input,
+			pending_focused_surface_destroy);
+
+	text_input->pending_focused_surface = NULL;
+	wl_list_remove(&text_input->pending_focused_surface_destroy.link);
+	wl_list_init(&text_input->pending_focused_surface_destroy.link);
+}
+
+static void
+relay_handle_new_text_input(struct wl_listener *listener, void *data)
```
JimmyCozza, line 524: If the surface already has keyboard focus when this is created, what sends enter? Sway checks the focused surface here.

input_method.c
```diff
@@ -0,0 +1,684 @@
...
+			.width = popup->focused_surface->current.width,
+			.height = popup->focused_surface->current.height,
+		};
+	}
+
+	popup_width = popup->popup_surface->surface->current.width;
+	popup_height = popup->popup_surface->surface->current.height;
+
+	/* Layout-space edges of the cursor rectangle */
+	x1 = parent_lx + cursor_area.x;
```
JimmyCozza, line 125: For XDG clients c->popups sits at the window-geometry origin, but cursor_rectangle is surface-local. Is this off by geometry.x/y for CSD clients?

input_method.c
```diff
@@ -0,0 +1,684 @@
...
+	x2 = x1 + cursor_area.width;
+	y1 = parent_ly + cursor_area.y;
+	y2 = y1 + cursor_area.height;
+
+	/* Default: left-aligned with the cursor, below it */
+	x = x1;
+	y = y2;
+
+	output = wlr_output_layout_output_at(output_layout, x1, y1);
```
JimmyCozza, line 134: popup_unconstrain in window.c uses the monitor workarea. Should this reuse it so the candidate window avoids the wibar?

input.c
```diff
@@ -1221,6 +1224,16 @@ keypress(struct wl_listener *listener, void *data)
 	if (some_is_lua_locked())
 		return;
 
+	/* An input method holding the keyboard grab consumes what the compositor
+	 * did not. Checked after keybindings so bindings still win over the IME. */
+	if ((kb_grab = input_method_get_keyboard_grab(group))) {
```
JimmyCozza, line 1229: handled only latches on presses. Does a release for a consumed keybinding still reach the grab, and then come back through the vk?

JimmyCozza, line 1229: also
Should this be gated on locked like the keybinding branch above?

input.c
```diff
@@ -1234,11 +1247,21 @@ keypressmod(struct wl_listener *listener, void *data)
...
-
-	wlr_seat_set_keyboard(seat, &group->wlr_group->keyboard);
-	/* Send modifiers to the client. */
-	wlr_seat_keyboard_notify_modifiers(seat,
-			&group->wlr_group->keyboard.modifiers);
+	struct wlr_input_method_keyboard_grab_v2 *kb_grab;
+
+	/* Modifiers follow the keys: while an input method holds the grab they go
+	 * to the IME rather than the focused client. */
+	if ((kb_grab = input_method_get_keyboard_grab(group))) {
```
JimmyCozza, line 1254: Same lock question as line 1229.

JimmyCozza (changes requested):
Thanks for the PR.  There are some things I'm trying to make sure I understand and think about.  

Could you also add some tests? Nothing in the suite speaks text-input-v3 or input-method-v2 yet, so this needs a small test client that can act as both a text input and a fake input method.

input_method.c
myf: it is indeed use after it was destroyed. i just copied the function from sway's and I will rewrite this in the next pull requestion. thank you

input_method.c
myf: this applies to wlroots 0.20 as well. but yes it should be stored at grab time instead. will update.

PR #777, opened 2026-09-12
fix(portal): ship a somewm-portals.conf

shuber2:
## Description

xdg-desktop-portal needs an explicit <desktop>-portals.conf to route most interfaces; somewm never shipped one. Commit f01b245 introduced somewm.portal, which declares Screenshot and ScreenCast but has no fallback for anything else, e.g., Firefox's dark/light theme detection (org.freedesktop.portal.Settings) doesn't work out of the box. It happened to work anyway for anyone who already had xdg-desktop-portal-gtk installed for an unrelated reason, since xdg-desktop-portal falls back to it as a last resort once nothing else claims an interface.

Add somewm-portals.conf, mirroring sway-portals.conf/wlroots-portals.conf: default to gtk, keep ScreenCast/Screenshot pinned to wlr.

## Validation

Before the fix, only the interfaces somewm.portal itself declares are exposed:

    $ busctl --user introspect org.freedesktop.portal.Desktop /org/freedesktop/portal/desktop | grep interface
    ...
    org.freedesktop.portal.ScreenCast
    org.freedesktop.portal.Screenshot
    ...
    (no Notification, Settings, FileChooser, etc.)

After the fix (with xdg-desktop-portal-gtk installed), restart the portal and recheck:

    $ systemctl --user restart xdg-desktop-portal.service
    $ busctl --user introspect org.freedesktop.portal.Desktop /org/freedesktop/portal/desktop | grep interface

Notification/Settings/FileChooser/etc. now appear.

## AI Usage

Used Claude Sonnet 5 when chasing a related bug, drafting the solution and validating with test scenarios on my system.


JimmyCozza (approved):
Looks good, I'll merge this tomorrow; if you want, add org.freedesktop.impl.portal.Inhibit=none` like sway does

shuber2:
You are very right on `org.freedesktop.impl.portal.Inhibit=none`.

## Fun background story on sway

On my Gentoo box, sway does not have such a line:
```
❯ cat /usr/share/xdg-desktop-portal/sway-portals.conf
[preferred]
# Use xdg-desktop-portal-gtk for every portal interface...
default=gtk
# ... except for the ScreenCast and Screenshot
org.freedesktop.impl.portal.ScreenCast=wlr
org.freedesktop.impl.portal.Screenshot=wlr
```

In fact, sway does not ship a conf file, but distributions do for them. Just did a bit more research, and there is a [github issue](https://github.com/swaywm/sway/issues/7773) on this from 2023. It says that sway would not officially support `xdg-desktop-portal`. But Arch then [committed](https://github.com/archlinux/aur/commit/c9727390d0032389d425ce4122f5869bd3c8e22e) a conf file. This, however, actually [looked like](https://github.com/archlinux/aur/commit/c9727390d0032389d425ce4122f5869bd3c8e22e#diff-90f87581e395bab9ce85c0bbc7dd9405eacb40261aa6e337c1d4e0add61823de) the one above.

JimmyCozza:
> In fact, sway does not ship a conf file, but distributions do for them

Oh, I see.  Yeah I think I had it backwards in my mind.

Thanks for this!

PR #779, opened 2026-09-12
fix: honor xdg_toplevel's requested fullscreen output

shuber2:
## Description

fullscreennotify() only read the fullscreen bool, never requested.fullscreen_output, so a client naming a specific target (e.g. wl-mirror --fullscreen-output) always ended up fullscreened on whatever monitor it already happened to be on instead.

## Validation

Launch `wl-mirror -S --fullscreen-output DP-9 eDP-1` on a terminal on eDP-1. Prior this patch the fullscreen window was shown on the screen that had focus (eDP-1), not on the declared DP-9.

## AI Usage

Tracked down but with Claude Sonnet. Manually investigated the guards implemented.


JimmyCozza:
@shuber2 

Does this handle a client that names an output before mapping? setmon() runs early, so mapnotify's setmon() early-returns and never applies the fullscreen state

I think this needs a client_surface(c)->mapped guard here plus reading requested.fullscreen_output where mapnotify picks target_mon. 

But maybe I'm wrong.

shuber2:
You are right. But since I did not have a client that would request `fullscreen_output` prematurely, which would trigger this situation, Claude generated a client for a test, and you were right. Generalized the commit. Is this what you had in mind?

JimmyCozza (approved):
Thanks!

PR #781, opened 2026-09-13
fix(awful.spawn): avoid O(n^2) complexity for output accumulation

JimmyCozza:
## Description
Cherry-pick of awesomeWM/awesome#4126 by @shuber2. 

`awful.spawn.easy_async` built its stdout and stderr strings by appending one line at a time, which copies the whole buffer on every line and gets slow fast on large output. 

The lines are now collected in a table and joined once, and the resulting strings are unchanged.


JimmyCozza:
We'll wait on this to see if it is accepted/changed in upstream Awesome.

Issue #782, opened 2026-09-14
GTask g_input_stream_real_read_async finalized without returning from awful.spawn.read_lines


mxmilkiib:
### Branch

`main` (2.0-dev)

### somewm --version output

```
somewm 2.0.0.r26.gcf53923.curscale-12
wlroots 0.18.x
Arch (Manjaro)
AMD Ryzen AI MAX+ PRO 395, Radeon 8060S
Wayland
```

### Problem

somewm logs recurring GLib warnings in pairs:

```
somewm[2163]: GTask g_input_stream_real_read_async (source object: 0x..., source tag: 0x...) finalized without ever returning (using g_task_return_*()). This potentially indicates a bug in the program.
```

22 occurrences over ~5 hours, always in pairs, 6-20 minutes apart.

### Root cause

`awful.spawn.read_lines` (`lib/awful/spawn.lua:525`) calls `stream:read_line_async()` without a `Gio.Cancellable`:

```lua
start_read = function()
    stream:read_line_async(GLib.PRIORITY_DEFAULT, nil, finish_read)
end
```

When a spawned process exits with empty stdout (e.g. `playerctl status` with no player, `brightnessctl` returning immediately), the pipe closes before the async read completes. GLib finalizes the GTask without calling `g_task_return_*()`, producing the warning.

The `done()` callback closes the stream but does not cancel the pending async read first, so GLib has no way to deliver a cancellation result to the pending GTask before teardown.

### Reproduction

1. Start somewm
2. Run a config that calls `awful.spawn.easy_async` on a command that exits with empty stdout on a timer (e.g. `playerctl status` every 5s with no player running)
3. Watch the journal: `journalctl --user -b | grep GTask`

### Expected behaviour

No GTask finalization warnings. The pending async read should be cancelled before the stream is closed.

### Proposed fix

Add a `Gio.Cancellable`, pass it to `read_line_async`, and cancel it in `done()` before closing the stream. Draft patch in the linked PR.

PR #783, opened 2026-09-14
fix(awful.spawn): cancel pending read_line_async before closing stream

mxmilkiib:
## Description

`awful.spawn.read_lines` calls `stream:read_line_async()` without a `Gio.Cancellable`. When a spawned process exits with empty stdout, the pipe closes before the async read completes, and GLib finalizes the GTask without `g_task_return_*()`, producing:

```
GTask g_input_stream_real_read_async finalized without ever returning
```

This addresses #782.

Fix: add a `Gio.Cancellable`, pass it to `read_line_async`, and cancel it in `done()` before closing the stream. This lets GLib deliver `g_task_return_cancelled` to the pending GTask, preventing the finalization warning.

The `finish_read` callback also checks `cancellable:is_cancelled()` and returns early if the cancellation fired, avoiding processing a result from a torn-down stream.

## Validation

- `luac -p` passes on the modified file
- Configs using `awful.spawn.easy_async` on a timer (`playerctl status`, `brightnessctl`, `powerprofilesctl get`) no longer produce GTask warnings in the journal over a 1-hour observation window
- Normal stdout-producing commands (`playerctl metadata` with an active player) still read lines correctly
- No regressions in `easy_async` callbacks receiving full stdout

Not yet verified: `with_line_callback` path (shares `read_lines` internally).

## AI Usage

Investigation and patch drafted with Devin (GLM-5.2 High). Root cause traced from journal warnings to the `read_line_async` call site; fix reviewed and syntax-verified by the author.


sicelo:
LGTM

JimmyCozza (commented):
Thanks @mxmilkiib 
A few questions:

- done() is only reached from finish_read, after read_line_finish has returned, so I think there is no pending read attached to the cancellable when cancel() runs on line 515.

  Am I reading that right? If so, is there a path where the cancel reaches anything?

- The is_cancelled branch on 529 only fires after done() has already run, and done() never starts another read, so I don't see how it is reachable.

  Is there a sequence I am missing?

- At 22 warnings in 5 hours, an hour would show about four, so a quiet hour is hard to tell apart from luck.

  Could you run the same config on an unpatched build for the same window?

- The warnings come in pairs, which matches one spawn's stdout and stderr, and both get a synchronous stream:close() from inside a GIO callback.

  Would it be worth checking whether that close, or the close_fd=true on GioUnix.InputStream.new, is what drops the task?

mxmilkiib:
You're right on both counts.

`done()` only runs after the read finished, so the `cancel()` there reaches nothing.

The `is_cancelled` branch can't fire either — dead code as written.

And the stats point stands: the warnings still fire with this patch loaded.

Every burst lands in the same second as a hot-reload — it was never about process exit.

The real mechanism: a pending read is parked via `g_task_attach_source`, so the GSource holds the task's last ref.

At reload, `luaA_cleanup_stale_glib_sources` destroys those sources, the task loses its last ref, and it finalizes without ever returning.

One read per pipe, hence pairs — six when three commands are mid-flight.

`close()` and `close_fd` are innocent; `done()` never runs because the state is already gone.

Pushed a rework: `read_lines` now registers each cancellable, and an "exit" handler cancels whatever is still in flight.

The reload drains the main context between "exit" and the sweep, so cancelled reads return inside the drain instead of being orphaned by it.

`is_cancelled()` does real work now — the cancelled callback fires in the drain, and that path closes the stream so the fd doesn't leak.

Verified nested: 9 warnings per reload before, 0 across three reloads after.

